PR #5, opened 2018-02-02
TextEncoder fix

haxxorsid:
I don't know what is the use of util.ts, but this is the fix to its error.

I am struggling to compile project properly in order to perform some changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
     "@types/react-icons": "^2.2.4",
     "@types/react-modal": "^3.1.1",
     "@types/react-split-pane": "^0.1.10",
-    "@types/text-encoding": "0.0.32",
     "@types/xterm": "^2.0.3",
     "cassowary": "0.0.2",
     "minimatch": "^3.0.4",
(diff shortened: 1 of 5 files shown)


mbebenita:
Thanks!

`TextEncoder` should be supported natively in Chrome and Firefox. The NPM package you included is a polyfill that's not necessary. You can probably just get away with declaring the TextEncoder as any to get around the type system and avoid having to take on a dependency on `text-encoding` module.

Are you getting any other compiler errors? I see there's one in `Test.tsx` that I should fix.

haxxorsid:
Yeah correct. But at least:
This should be replaced:
`var bytes = new TextEncoder('utf-8').encode(buffer);`
With this:
`var bytes = new TextDecoder('utf-8').decode(buffer);`
In order to solve: **TS2345**
Because we are converting ArrayBuffer to String, encode accepts only String.

haxxorsid:
For WasmParser error: wasdk/wasmparser@f9f3afb65eb3a74be8ee6b5e7f89768bd35b5677

mbebenita:
Oh, I just realized from your comment that base64Encode is not even being used yet, feel free to remove it for now, and we can add it back in later when we need it.

mbebenita:
I think this is no longer necessary, fixed by other commits. Closing this PR.

PR #9, opened 2018-02-03
Fix travis

AnshulMalik:
I removed some redundant code, we can add it later when actually required.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "webpack --config webpack.config.js",
     "build-watch": "webpack --watch --config webpack.config.js",
-    "dev-server": "webpack-dev-server --config webpack.config.js"
+    "dev-server": "webpack-dev-server --config webpack.config.js",
+    "clean": "rm -r dist/*"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


nileshgulia1:
Thanks @AnshulMalik  for changes, assuming that `react-split-pane` not needed right now, may be @mbebenita  can review it ,Otherwise good work :)

.travis.yml
```diff
@@ -13,12 +13,7 @@ notifications:
   email:
     on_success: never
     on_failure: always
-before_install:
-- npm update
 install:
 - npm install
-before_script:
-- npm run clean
```
nileshgulia1: May be there is no need of removing clean script ,we can add this to `package.json`:
`"clean": "rm -r dist/*"` what it does is just removes a bunch of files so you can start with a fresh working copy to start building into.

mbebenita:
> assuming that react-split-pane not needed right now

You are correct that `react-split-pane` is not needed and can be removed.

mbebenita:
@AnshulMalik when submitting PRs, please don't include merge commits because it makes it difficulty to review code. Usually, you'll want to rely on git rebase to have all your commits in a linear history rebased on the current master.

mbebenita:
Looks good.

PR #10, opened 2018-02-03
Exporting to Gist

haxxorsid:
Exports fiddle to Gist


src/service.ts
```diff
@@ -308,7 +324,23 @@ export class Service {
     }
     return uri;
   }
-
+  static async exportProject(project: Project, uri?: string): Promise<string> {
+    let files: any = {};
+    function serialize(file: File) {
```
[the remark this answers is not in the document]
haxxorsid: No.[Gist API]( https://developer.github.com/v3/gists/#create-a-gist )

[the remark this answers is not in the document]
AnshulMalik: One more thing, the directory structure is not preserved here

src/service.ts
```diff
@@ -308,7 +324,23 @@ export class Service {
...
+      if (file instanceof Directory) {
+        file.mapEachFile((file: File) => serialize(file));
+      } else {
+        files[file.name] = {content: file.data};
+      }
+    }
+    let json: any = { description: "source: http://webassembly.studio", public: true, files: {}};
+    serialize(project);
+    json.files = files;
+    if(!isUndefined(uri))
```
AnshulMalik: I can't find this function anywhere :P `isUndefined`

AnshulMalik:
In case of built files, the generated `.wasm` file's content is not a string, instead it's an object

screen shot 2018-02-04 at 12 01 50 am


So the gist upload fails, complaining that it wants a string.

haxxorsid:
1) Gist doesn't support directories
2) Undefined is part of @types/node
3) For saving build: it is showing 422, will need a change.

AnshulMalik:
Aren't we using some chat platform to discuss things?

src/components/Toast.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from "react";
+import * as ReactModal from "react-modal";
+import { Button } from "./Button";
+import { GoX } from "./Icons";
+
+export class Toast extends React.Component<{
```
mbebenita: The `Toast` component should be generic so that we can use it in other cases. The current implementation is specific to gists. Can you please rename the `uri` property to `message` and change the `render` function to reflect this. The "Gust Created ..." message can be passed in as a property.

src/components/Toast.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    render() {
+
+        return <ReactModal
+            isOpen={true}
+            contentLabel="Gist Created"
+            className="toast"
+            ariaHideApp={false}
+            style={{overlay: {background: "rgba(255, 255, 255, 0)", position: "inherit"}}}>
+            <div style={{ flexDirection: "column", height: "100%" }}>
```
mbebenita: CSS styling should be moved into a CSS file.

mbebenita:
> chat platform

At Mozilla we generally use IRC. We spend time in the #wasm channel on the irc.mozilla.org network.
I'm open to other platforms, like slack, discourse, etc. if anyone wants to set things up.

src/service.ts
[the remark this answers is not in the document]
AnshulMalik: It'd be good to add a new file which would contain the directory structure, it'd be easier to reproduce the whole project from a gist then.

AnshulMalik:
I have created a slack app, and setup an invite bot https://wasm-studio-invite.herokuapp.com/ 🥂

haxxorsid:
@mbebenita No, the gist doesn't support directory structure.

src/service.ts
```diff
@@ -308,7 +324,23 @@ export class Service {
...
+      json["description"] = json["description"] + `/?f=${uri}`;
```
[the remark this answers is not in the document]
haxxorsid: I tried to make it refer to the original fiddle's URL(in case it is forked). Like JSBin does after exporting a saved fiddle to gist.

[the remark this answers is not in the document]
haxxorsid: I tried to reflect the fiddle's URI in case it is forked, just like JSBin does.

haxxorsid:
I have made changes as asked. Adding "generated" flag to Directory/File is remaining.

src/components/Toast.tsx
haxxorsid: Done.

src/components/App.tsx
```diff
@@ -620,7 +620,9 @@ export class App extends React.Component<AppProps, AppState> {
 
     return <div className="fill">
       {this.state.toast &&
-          <Toast onDismiss={this.dismissToast.bind(this)} uri={this.state.toast}/>
+        <Toast onDismiss={this.dismissToast.bind(this)}
+               message={["Gist Created!", <a href={this.state.toast} target='_blank' className='gist-link'>Open in new tab.</a>]}
```
mbebenita: Instead of passing an array for message parameter, you can pass in a JSX element. This could become simply:

```
message={<div>Gist Created! ... </div>}
```

src/components/Button.tsx
```diff
@@ -6,12 +6,12 @@ export class Button extends React.Component<{
...
   isDisabled?: boolean;
   onClick?: Function;
-  customClass?: string;
+  customClassName?: string;
 }, {}> {
   render() {
     let className = "button";
-    if(this.props.customClass){
-        className = this.props.customClass
+    if(this.props.customClassName){
```
mbebenita: Formatting nit, space after `if`, also before the brace. Now that we just landed tslint  this will be automatically detected in the future.

src/components/Button.tsx
```diff
@@ -6,12 +6,12 @@ export class Button extends React.Component<{
...
+      className += this.props.customClassName
```
mbebenita: It's better to insert the space here `className += " " + this.props.customClassName` than rely on doing so at each call site. Also, a missing semicolon.

src/components/Toast.tsx
```diff
@@ -4,30 +4,28 @@ import { Button } from "./Button";
 import { GoX } from "./Icons";
 
 export class Toast extends React.Component<{
-    onDismiss: Function;
-    uri: string;
+  onDismiss: Function;
+  message: Array<any>;
```
mbebenita: This can become `message: JSX.Element`;

haxxorsid:
I am deleting branch closing this PR because of some rebasing issue.
I will create another PR.

PR #11, opened 2018-02-03
Add tslint

AnshulMalik:
Linting is a necessary step in the process, it'll be good if we enforce some coding guidelines since codebase will keep on growing and then it'd be harder.

This PR only contains changes related to code style.


package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "webpack --config webpack.config.js",
     "build-watch": "webpack --watch --config webpack.config.js",
-    "dev-server": "webpack-dev-server --config webpack.config.js"
+    "dev-server": "webpack-dev-server --config webpack.config.js",
+    "test": "./node_modules/.bin/tslint -p ."
```
nileshgulia1: @AnshulMalik  Please change this to `"tslint": ""./node_modules/.bin/tslint -p ."` as unit and integration test scripts will come here,otherwise conflicts will be there.

mbebenita:
This is great. Can you please rebase and squash all the changes into one commit, this would make it much easier to review.

tslint.json
```diff
@@ -0,0 +1,29 @@
+{
+    "defaultSeverity": "error",
+    "extends": [
+        "tslint:recommended"
+    ],
+    "exclude": [ "./src/languages/*.ts"],
```
mbebenita: What are the reasons for excluding the files in the `languages` directory.

mbebenita: Also, I think you should exclude the `lib` directory since that code is not part of this project.

AnshulMalik:
Looks like I messed up :P, working on fix

AnshulMalik:
We need to merge #9 before this

AnshulMalik:
@mbebenita @nileshgulia1  Done the changes. But I was wondering how merge commits hinder in the review process?

mbebenita:
> But I was wondering how merge commits hinder in the review process?

As a reviewer, I'm only interested in the changes that are going to be applied on master. Merge commits are a detail of your workflow.

mbebenita:
👍  super!!

PR #14, opened 2018-02-04
Exporting to Gist

haxxorsid:
Recreated PR: #10  because of some confusing rebase.

Fixed linting issues and readded all the previous commits. Ready to land.


src/components/Icons.tsx
```diff
@@ -107,4 +107,10 @@ export class GoThreeBars extends React.PureComponent {
   render() {
     return <svg className="octicon octicon-three-bars" viewBox="0 0 12 16" version="1.1" width="12" height="16" aria-hidden="true"><path fillRule="evenodd" d="M11.41 9H.59C0 9 0 8.59 0 8c0-.59 0-1 .59-1H11.4c.59 0 .59.41.59 1 0 .59 0 1-.59 1h.01zm0-4H.59C0 5 0 4.59 0 4c0-.59 0-1 .59-1H11.4c.59 0 .59.41.59 1 0 .59 0 1-.59 1h.01zM.59 11H11.4c.59 0 .59.41.59 1 0 .59 0 1-.59 1H.59C0 13 0 12.59 0 12c0-.59 0-1 .59-1z"></path></svg>
   }
+}
+
+export class GoGist extends React.PureComponent {
+    render() {
+        return <svg className="octicon octicon-gist" viewBox="0 0 12 16" version="1.1" width="12" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M7.5 5L10 7.5 7.5 10l-.75-.75L8.5 7.5 6.75 5.75 7.5 5zm-3 0L2 7.5 4.5 10l.75-.75L3.5 7.5l1.75-1.75L4.5 5zM0 13V2c0-.55.45-1 1-1h10c.55 0 1 .45 1 1v11c0 .55-.45 1-1 1H1c-.55 0-1-.45-1-1zm1 0h10V2H1v11z"></path></svg>
+    }
 }
```
nileshgulia1: Make sure you place a newline at the end.

haxxorsid: It is showing 117 lines in my git. I have pushed all the changes, I wonder why it is 116 here. I will fix it.

haxxorsid:
@nileshgulia1 That is because of my previous commits. I haven't repeated any commit on package-lock.json. You can just ignore that.

src/components/App.tsx
```diff
@@ -412,6 +419,26 @@ export class App extends React.Component<AppProps, AppState> {
...
+        this.setState({
+            toast:  gistURI
+        });
+        console.log(`Gist created: ${gistURI}`);
+    } else {
+        console.log("Failed!");
+    }
+  }
+
+  dismissToast(index: number) {
```
AnshulMalik: Shouldn't this be `onDismissToast`? `dismissToast` conveys different message, that I need to call this method to dismiss the toast

src/components/App.tsx
```diff
@@ -123,6 +124,11 @@ export interface AppState {
    */
   editorSplits: SplitInfo[];
 
+  /**
```
AnshulMalik: This comment is unnecessary :P

src/components/App.tsx
```diff
@@ -588,6 +618,11 @@ export class App extends React.Component<AppProps, AppState> {
     </Split>;
 
     return <div className="fill">
+      {this.state.toast &&
```
AnshulMalik: I vote for using
```
if(this.state.toast) {
  return <Toast />
}
```
Or something like that. `&&` looks confusing.

src/components/Button.tsx
```diff
@@ -5,10 +5,14 @@ export class Button extends React.Component<{
...
   title?: string;
   isDisabled?: boolean;
-  onClick?: Function
+  onClick?: Function;
+  customClassName?: string;
 }, {}> {
   render() {
     let className = "button";
+    if (this.props.customClassName) {
+      className += this.props.customClassName;
```
AnshulMalik: As @mbebenita said, there should be a space in between classes

src/components/App.tsx
```diff
@@ -412,6 +419,26 @@ export class App extends React.Component<AppProps, AppState> {
     }
     this.setState({ fiddle });
   }
+  async gist() {
```
AnshulMalik: Can we name it to something more readable?

src/components/Button.tsx
haxxorsid: Add that in linting. It doesn't show me any issue.

src/components/App.tsx
mbebenita: Using `&&` is the JSX way of doing this.

mbebenita:
@haxxorsid I manually merged your changes and addressed the last comment I made. There is some follow up work on this but I'll open an issue for it. Thanks!

PR #21, opened 2018-02-05
add package-lock.json to gitignore

nileshgulia1:
We don't need to commit this everytime, this will allow us to always test the latest version of our dependencies especially in CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 dist
 node_modules
+package-lock.json


AnshulMalik:
I believe package-lock should be kept in the repo. https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

yurydelendik:
> I believe package-lock should be kept in the repo.

I agree. There are two lines of thoughts: keep package-lock.json or not. In our case, it will be beneficial to keep entire tool set and referred packages on the same version on every computer to reproduce the same built code. That said, any change to the package-lock.json shall be considered as library/dependency upgrade, therefore the change deserves a separate PR IHMO

mbebenita:
I agree with @yurydelendik, any time we upgrade our dependencies we need to test manually.

nileshgulia1:
@AnshulMalik  @yurydelendik  @mbebenita  Its better closing it then :)

Issue #19, opened 2018-02-05
Stacked Toasts


mbebenita:
We should work out a scheme where more toasts can be fired off and stacked up on top of each other.

AnshulMalik:
Should we use a package for this or build our own?

AnshulMalik:
If we are to use external package, https://github.com/fkhadra/react-toastify is a good candidate

haxxorsid:
People can refer to this: [wasdk/WasmFiddle#11](https://github.com/wasdk/WasmFiddle/pull/11)

AnshulMalik:
Looks awesome @haxxorsid, waiting to see the PR here.

mbebenita:
I've tried this component before, and found it difficult to style. In my experience, rolling our own components that do exactly what we want them to do is easier than depending on another codebase that does almost what we want, and a lot of things we don't.

haxxorsid:
Toasts are using React-Modal. React-modal doesn't allow a single wrapper wrapping multiple modals. And we need a wrapper around all modals to make them stackable. So we will have to remove react-modal and make a pure toast just like [wasdk/WasmFiddle#11](https://github.com/wasdk/WasmFiddle/pull/11)

mbebenita:
I'm open to using a more advanced Toast implementation.

haxxorsid:
This issue can be closed.

haxxorsid:
This issue can be closed.

Issue #18, opened 2018-02-05
Better Toast Placement


mbebenita:
The current Toast placement is hard to see:

![image](https://user-images.githubusercontent.com/311082/35783928-4fa8ecf8-09c5-11e8-9136-874858a10153.png)

Could we put this at the bottom? Somewhere more obvious?

haxxorsid:
Toasts have been position in the bottom. This issue can be closed.

PR #24, opened 2018-02-05
Stackable and Toast positioned at Bottom

haxxorsid:
1. Positioned toast at bottom
2. Stackable multiple toasts
Addresses: [#10](https://github.com/wasdk/WebAssemblyStudio/issues/19), [#18](https://github.com/wasdk/WebAssemblyStudio/issues/18)


src/components/App.tsx
```diff
@@ -422,19 +417,19 @@ export class App extends React.Component<AppProps, AppState> {
...
       console.log("Failed!");
     }
   }
 
   onToastDismiss(index: number) {
-      this.setState({
-          toast:  null
-      });
+    this.setState((prevState: any) => ({
+      toasts: prevState.toasts.filter((key: number, value: number) => value !== index )
```
AnshulMalik: I think the `Toast` component should have the toasts in its state, or in redux store(do we have it?)

haxxorsid: 1) Action/Event is in App(Parent), so the state is kept in App.
2) No redux.

AnshulMalik: Yeah, what I was saying is, it should be independent since we might have many components who would like to generate toasts.

mbebenita: I think it's okay to have the Toast be a controlled component. We can write a ToastContainer component that wraps it and provides an immediate mode API, like `showToast(...)`; The App component can hold a reference to the ToastContainer. What do you think?

src/components/App.tsx
```diff
@@ -666,13 +661,15 @@ export class App extends React.Component<AppProps, AppState> {
     </Split>;
 
     return <div className="fill">
-      {this.state.toast &&
+      <div className="toast-container">
+      {this.state.toasts && this.state.toasts.map((toast: string, key: number) =>
```
mbebenita: What about other kinds of toasts? I think of toasts as notifying the user of events that they may need to take an action on. This seems hardcoded to Gist toasts.

mbebenita: Also, you probably want to check the length of the toasts, not just the presence of the array.

src/components/App.tsx
```diff
@@ -666,13 +661,15 @@ export class App extends React.Component<AppProps, AppState> {
     </Split>;
 
     return <div className="fill">
-      {this.state.toast &&
+      <div className="toast-container">
```
mbebenita: This feels like it shouldn't belong in the App component, it's an internal detail of how toasts are displayed. Perhaps you need a `ToastContainer` component that accepts `Toast` elements as children.

haxxorsid:
Conflicts removed, but Toast.tsx is redundant.

src/components/App.tsx
```diff
@@ -163,10 +163,10 @@ export class App extends React.Component<AppProps, AppState> {
         { min: 40, value: 256 }
       ],
       editorSplits: [],
-      toasts: [],
       showProblems: true,
       showSandbox: true
     };
+    this.toastInstance = new ToastContainer();
```
mbebenita: You should not create the instance here. The ref assignment does this for you in the render() function.

mbebenita:
@haxxorsid Can you please squash all your commits into one and force-push? This makes it a lot easier to review this PR.

src/components/Toasts.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from "react";
```
yurydelendik: License header is missing

haxxorsid:
@mbebenita Sure.

haxxorsid:
@mbebenita Done. All squashed.
Can you yourself commit( after accepting this PR) this:  remove toast.tsx and remove that `this.toastInstance = new ToastContainer(). I am now done for today.

src/components/App.tsx
```diff
@@ -422,21 +417,13 @@ export class App extends React.Component<AppProps, AppState> {
     const gistURI = await Service.exportProjectToGist(this.project, this.state.fiddle);
     this.logLn("Project Gist CREATED ");
     if (gistURI) {
-      this.setState({
-        toast:  gistURI
-      });
+      this.toastInstance.showToast(<span>"Gist Created!" <a href={gistURI} target="_blank" className="toast-span">Open in new tab.</a></span>);
```
mbebenita: This is okay for now, but I think it's a React anti-pattern. I'm not a React expert by any means, but I think elements should only be created in render functions.

mbebenita:
> yourself commit

@haxxorsid sure, I can do that. Thanks.

mbebenita:
Code in this PR has been merged via #31, closing.

Issue #23, opened 2018-02-05
Move service endpoint into external configuration file


yurydelendik:
We use couple of external services. It's recommended to move the endpoints addressed (e.g. `//wasmexplorer-service.herokuapp.com/service.php`) into some configuration file and don't include it into the bundle.

yurydelendik:
Affected places:
- rustc https://github.com/wasdk/WebAssemblyStudio/blob/master/src/service.ts#L198
- other https://github.com/wasdk/WebAssemblyStudio/blob/master/src/service.ts#L98

yurydelendik:
Closed via #29

PR #29, opened 2018-02-06
Move service-url to config file

AnshulMalik:


yurydelendik:
Fixes #23

It is still part of the bundle. Can we place it in the external file and read it using fetch()?

Sorry, I also did not mention is the issue, there is also service for the rust. Can you add this config to the PR?

AnshulMalik:
Where should the external file exist? 

Don't know why this doesn't look right to me

yurydelendik:
> Where should the external file exist?

Let's say we will name it e.g. "wasmservices.config" and place in the same folder as "index.html", e.g. in JSON format. The ts code will read this file asynchronously, probably cache, and give service address to the service.ts (via Promise/async).

config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "serviceUrl": "http://wasmexplorer-service.herokuapp.com/service.php",
```
yurydelendik: start URL with '//' to not specify a schema

(you can modify the one below too)

src/service.ts
```diff
@@ -93,12 +93,27 @@ export interface IServiceRequest {
...
+}
+
 export class Service {
-  static async sendRequest(command: string): Promise<IServiceRequest> {
-    const response = await fetch("//wasmexplorer-service.herokuapp.com/service.php", {
+  static config: IConfig;
+
+  static async sendRequest(body: string, to: string): Promise<IServiceRequest> {
+    if (!this.config) {
+      const configUrl = `${location.protocol}//${location.hostname}${(location.port ? ":" + location.port : "")}/config.json`;
```
yurydelendik: define `configUrl` as './config.json'

src/service.ts
```diff
@@ -93,12 +93,27 @@ export interface IServiceRequest {
...
+      this.config = await fetch(configUrl).then(resp => resp.json());
```
yurydelendik: Extract `IConfig` and `fetch(configUrl).then(resp => resp.json())` into separate .ts module.

src/service.ts
```diff
@@ -189,19 +204,13 @@ export class Service {
...
     } else if (from === Language.Rust && to === Language.Wasm) {
       // TODO: Temporary until we integrate rustc into the service.
-      return fetch("https://rust-heroku.herokuapp.com/rustc", {
-        method: "POST",
-        body: JSON.stringify({code: src}),
-        headers: new Headers({
-          "Content-Type": "application/json"
-        })
-      }).then(res => res.json());
+      return this.sendRequest(JSON.stringify({code: src}), "rustc");
```
yurydelendik: It's hard to follow what `sendRequest` returns, in this case an object expected, however it was changed to return string.

Can you:
 - remove inaccurate refactoring and keep it as it was before
 - or, introduce sendRequestJSON to handle a new way of communicating with service?

src/service.ts
```diff
@@ -189,19 +204,13 @@ export class Service {
         ]
       };
       const input = encodeURIComponent(JSON.stringify(project)).replace("%20", "+");
-      return this.sendRequest("input=" + input + "&action=build");
+      return this.sendRequest("input=" + input + "&action=build", "service");
```
yurydelendik: Instead of passing "rustc" or "service" as a string, please introduce an enumeration.

src/service.ts
AnshulMalik: `sendRequest` return `return JSON.parse(await response.text());` which will be an object

src/config.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+const configUrl = "./config.json";
+
+export default async function getConfig() {
+  if (!this.config) {
+    this.config = await fetch(configUrl).then(resp => resp.json());
+  }
+
+  return this.config;
```
yurydelendik: what is `this` in this context? for a module, you can declare the variable above

src/service.ts
```diff
@@ -93,12 +94,32 @@ export interface IServiceRequest {
...
+
+    if (to === ServiceTypes.rustc) {
+      body = JSON.stringify(content);
+      url = config.rustc;
+      contentType = "application/json";
+    } else {
+      body = content.toString();
+      url = config.serviceUrl;
+      contentType = "application/x-www-form-urlencoded";
+    }
```
yurydelendik: It's hard to the code logic. Could you create two methods `static async sendRequest(content: string | Object, to: ServiceTypes): Promise<IServiceRequest>` and `static async sendRequestJSON(content: string | Object, to: ServiceTypes): Promise<IServiceRequest>` to remove logic around assigning body/url/contentType variables and reduce complexity for simple method?

AnshulMalik: will do

src/service.ts
```diff
@@ -93,14 +94,34 @@ export interface IServiceRequest {
   output: string;
 }
 
+export enum ServiceTypes {
+  rustc,
```
yurydelendik: nit: make enumeration values in Pascal case.

src/service.ts
```diff
@@ -93,14 +94,34 @@ export interface IServiceRequest {
...
+      headers: new Headers({ "Content-Type": "application/json" })
     });
-    return JSON.parse(await response.text());
+
+    return response.json();
+  }
+
+  static async sendRequest(content: string, to: ServiceTypes): Promise<IServiceRequest> {
+    const config = await getConfig();
+    const url = (to === ServiceTypes.rustc) ? config.rustc : config.serviceUrl;
```
yurydelendik: nit: no need for parens `()` in this expression

src/config.ts
```diff
@@ -0,0 +1,15 @@
+export interface IConfig {
```
yurydelendik: License header is missing

yurydelendik:
Thank you for the patch

PR #36, opened 2018-02-07
Uploading files

haxxorsid:
### Summary of Changes

* Added Upload file dialog.
* User can upload multiple files, all selected files are listed in ListBox
* User can see errors related file upload(eg: file already exists, unsupported ext) in the upload-file-dialog.

### Improvements needed
* UploadFileDialog.tsx can be improved.
* Possible improvements are stated below in reviews.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -143,6 +143,9 @@ export interface AppState {
    * Editor split state.
    */
   editorSplits: SplitInfo[];
+  /**
+   * If true, the upload file dialog is open.
+   */
   uploadFileDialogDirectory: Directory;
   showProblems: boolean;
   showSandbox: boolean;
@@ -743,7 +746,7 @@ export class App extends React.Component<AppProps, AppState> {
           onUpload={(files: File[]) => {
             if (!(this.state.uploadFileDialogDirectory instanceof Project)) {
               files.map((file: File) => {
-                this.state.uploadFileDialogDirectory.children.push(file);
+                this.state.uploadFileDialogDirectory.addFile(file);
               });
             } else {
               files.map((file: File) => {
@@ -796,7 +799,7 @@ export class App extends React.Component<AppProps, AppState> {
               this.forceUpdate();
             }}
             onUploadFile={(directory: Directory) => {
-                this.setState({ uploadFileDialogDirectory: directory});
+              this.setState({ uploadFileDialogDirectory: directory});
             }}
           />
           <div className="fill">
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -741,10 +741,7 @@ export class App extends React.Component<AppProps, AppState> {
             this.setState({ uploadFileDialogDirectory: null });
            }}
           onUpload={(files: File[]) => {
-            console.log(this.state.uploadFileDialogDirectory);
-            console.log(this.project);
             if (!(this.state.uploadFileDialogDirectory instanceof Project)) {
-              console.log("true");
               files.map((file: File) => {
                 this.state.uploadFileDialogDirectory.children.push(file);
               });
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -45,6 +45,7 @@ import { Button } from "./Button";
 import * as ReactModal from "react-modal";
 import { NewFileDialog } from "./NewFileDialog";
 import { EditFileDialog } from "./EditFileDialog";
+import { UploadFileDialog } from "./UploadFileDialog";
 import { ToastContainer } from "./Toasts";
 import { Spacer, Divider } from "./Widgets";
 import { Cton } from "../languages/cton";
@@ -142,6 +143,7 @@ export interface AppState {
    * Editor split state.
    */
   editorSplits: SplitInfo[];
+  uploadFileDialogDirectory: Directory;
   showProblems: boolean;
   showSandbox: boolean;
 }
@@ -185,7 +187,8 @@ export class App extends React.Component<AppProps, AppState> {
       ],
       editorSplits: [],
       showProblems: true,
-      showSandbox: true
+      showSandbox: true,
+      uploadFileDialogDirectory: null
     };
     this.registerLanguages();
   }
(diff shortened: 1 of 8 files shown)


src/components/UploadFileDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+      const name: string = file.name;
+      const fileType: string = filetypeForExtension(name.split(".").pop());
+      let data: any;
+      try {
+        data = await this._readUploadedFileAsText(file);
+        this.setState(prevState => ({
+          files: [...prevState.files, {name, fileType, data}]
+        }));
+      } catch (e) {
+          console.log("Unable to read the file!");
```
haxxorsid: This needs to be handled like stuff in _nameError(). Idk how should I connect this with output of _nameError.

src/components/UploadFileDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+  async _handleUpload(files: FileList) {
+    this.setState({files: []});
+    Array.from(files).forEach(async (file: any) => {
+      const name: string = file.name;
+      const fileType: string = filetypeForExtension(name.split(".").pop());
+      let data: any;
+      try {
+        data = await this._readUploadedFileAsText(file);
+        this.setState(prevState => ({
+          files: [...prevState.files, {name, fileType, data}]
```
haxxorsid: `data `of file is added in state even if _nameError function gives some errors. So data must be added later. I need solution for adding data in state later.

src/components/App.tsx
```diff
@@ -730,6 +733,27 @@ export class App extends React.Component<AppProps, AppState> {
...
+        <UploadFileDialog
+          isOpen={true}
+          directory={this.state.uploadFileDialogDirectory}
+          onCancel={() => {
+            this.setState({ uploadFileDialogDirectory: null });
+           }}
+          onUpload={(files: File[]) => {
+            if (!(this.state.uploadFileDialogDirectory instanceof Project)) {
+              files.map((file: File) => {
+                this.state.uploadFileDialogDirectory.children.push(file);
```
haxxorsid: Oh. Maybe I should use addFile function here instead of pushing?

src/components/UploadFileDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+    this.state = {
+      files: []
+    };
+  }
+  _nameError() {
+    const directory = this.props.directory;
+    let errorStr: string = "";
+    this.state.files.map((file: any) => {
+      if (file.fileType == null) {
+          errorStr += `File '${file.name}' is not supported.`;
```
mbebenita: Indentation.

src/components/UploadFileDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+              this.props.onCancel();
+            }}
+          />
+          <Button
+            icon={<GoFile />}
+            label="Upload"
+            title="Upload New File"
+            isDisabled={!this.state.files.length || !!this._nameError()}
+            onClick={() => {
+            const newFiles: File[] = [];
```
mbebenita: Indentation.

src/components/UploadFileDialog.tsx
mbebenita: Can you associate errors with each file structure in `state.files`, and maybe show errors inside of the list items?

mbebenita:
@yurydelendik is it possible to upload directories?

mbebenita:
@haxxorsid instead showing files using a list box, we could embed a DirectoryTree view in the upload file dialog. Instead of keeping track of files as an array of files, you could create a Directory and stick the files in there, and that's the object you'd pass to the embedded DirectoryTree. To prevent users from right clicking on files in the embedded DirectoryTree component, we can add a prop to it that disables that functionality.

mbebenita:
Lastly, we can probably add a button at the bottom after Cancel that is styled like the other buttons, which clicks on the upload input element.

image

yurydelendik:
> @yurydelendik is it possible to upload directories?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement lists `allowdirs` and `webkitdirectory` properties -- I guess it's possible

src/components/UploadFileDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+  }
+  _nameError() {
+    const directory = this.props.directory;
+    let errorStr: string = "";
+    this.state.files.map((file: any) => {
+      if (file.fileType == null) {
+          errorStr += `File '${file.name}' is not supported.`;
+      }
+      if (file.name) {
+        if (!/^[a-z0-9\.\-\_]+$/i.test(file.name)) {
```
yurydelendik: Extract regexp check as a function with meaningful name, e.g. `isFilenameValid`

src/components/UploadFileDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+      } catch (e) {
+          console.log("Unable to read the file!");
+      }
+    });
+  }
+  _readUploadedFileAsText(inputFile: any) {
+    const temporaryFileReader = new FileReader();
+    return new Promise((resolve, reject) => {
+      temporaryFileReader.onerror = () => {
+      temporaryFileReader.abort();
```
yurydelendik: nit: bad indent

mbebenita:
> I guess it's possible

Cool, so this means we could recursively upload them and display them into a DirectoryTree component, maybe with some indicators like (this file already exists, etc).

mbebenita:
@haxxorsid let's handle the non directory case for now, and switch to the DirectoryTree view, we can do the rest later.

src/components/UploadFileDialog.tsx
mbebenita: Not sure why tslint is not reporting these errors.

haxxorsid:
"Lastly, we can probably add a button at the bottom after Cancel that is styled like the other buttons, which clicks on the upload input element." 
You mean it will open browser's upload file dialog?

mbebenita:
> it will open browser's upload file dialog

Yes, that's possible. @yurydelendik has some tricks.

haxxorsid:
@mbebenita 
Assume: I have this <input id='uploadfile' type='file' multiple hidden/>
I do this in Button Component:
```
onClick={() => {
  document.getElementById('uploadfile').click();
}}
```
Cool?

mbebenita:
For some reason I can't see this comment in github, so replying over
e-mail. But yes, you should be able to do that, and then hide the
uploadfile element with CSS.

src/components/UploadFileDialog.tsx
```diff
@@ -128,16 +137,24 @@ export class UploadFileDialog extends React.Component<UploadFileDialogProps, {
           <Button
             icon={<GoFile />}
             label="Upload"
-            title="Upload New File"
-            isDisabled={!this.state.files.length || !!this._nameError()}
+            title="Upload files"
+            onClick={() => {
+              document.getElementById("file-upload-input").click();
```
mbebenita: This is a React anti-pattern. I think what you need to do, is add a `ref` on the FileUploadInput component, then here you can say something like `this.fileUploadInput.open()`.

src/components/Widgets.tsx
```diff
@@ -85,22 +84,7 @@ export class FileUploadInput extends React.Component<{
...
-            <div className="column">
-              <div>{input}</div>
-            </div>
-          </div>
-          {this.props.error && <div className="text-input-box-error">{this.props.error}</div>}
-        </div>;
-    } else {
-      return input;
-    }
+    return <input id="file-upload-input" type="file" onChange={this.props.onChange} multiple hidden/>;
```
mbebenita: Here you can do a `ref` to get a handle to the input element. Then add a function called `open()` that files the click event on it, which is called from a parent component.

haxxorsid:
@mbebenita Yep.

haxxorsid:
@mbebenita I've fixed everything including indents. Euhm I don't know what is blocking this more.

mbebenita:
@haxxorsid thanks!

PR #45, opened 2018-02-08
Download project

haxxorsid:
### Summary of Changes

* Added download button.
* Used JSZip for creating zipfile and it's blob.


src/service.ts
```diff
@@ -331,6 +332,37 @@ export class Service {
...
+        } else {
+          zipFile.file(file.name, file.data);
+        }
+      }
+    }
+    defineZip(project);
+    zipFile.generateAsync({type: "base64"}).then(function(base64: any) {
+      const link = document.createElement("a");
+      link.download = zipName;
+      link.href = "data:application/zip;name=test.zip;base64," + base64;
```
yurydelendik: there is can be a limit on size of the URL. I recommend to use ['blob' type](https://stuk.github.io/jszip/documentation/api_jszip/generate_async.html#type-option) with generateAsync

haxxorsid: It will need FileSaver.js to provide zip as a download to the user. So basically it will add up more size in the bundle. Is there a way to provide zip through blob(without adding up more libs)?

yurydelendik: Sorry, I don't understand relation between FileSaver.js. URL.createObjectURL gives blob to be use in a web app.

package.json
```diff
@@ -39,12 +39,16 @@
...
   "dependencies": {
+    "@types/file-saver": "^1.3.0",
+    "@types/jszip": "^3.1.3",
     "@types/minimatch": "^3.0.2",
     "@types/mousetrap": "^1.5.34",
     "@types/react-icons": "^2.2.4",
     "@types/react-modal": "^3.1.1",
     "@types/xterm": "^2.0.3",
     "cassowary": "0.0.2",
+    "file-saver": "^1.3.3",
```
yurydelendik: Why do we need file-saver?

haxxorsid: For: [this](https://github.com/haxxorsid/WebAssemblyStudio/blob/0cd09ac9131edf67ddabb7bfb75b8ebf5c08bcbb/src/service.ts#L360)

src/service.ts
```diff
@@ -331,6 +333,34 @@ export class Service {
...
+      } else {
+        if (!isUndefined(folder)) {
+          folder.file(file.name, file.data);
+        } else {
+          zipFile.file(file.name, file.data);
+        }
+      }
+    }
+    defineZip(project);
+    zipFile.generateAsync({type: "blob"}).then(function(blob: any) {
```
yurydelendik: nit: use arrow function instead

package.json
yurydelendik: The alternative for generating URL from blob is [URL.createObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL)

src/service.ts
yurydelendik: see also https://stuk.github.io/jszip/documentation/api_jszip/generate_async.html#mimetype-option example how to specify a mime type for generated blob.

src/service.ts
```diff
@@ -331,6 +332,37 @@ export class Service {
...
+          zipFile.file(file.name, file.data);
+        }
+      }
+    }
+    defineZip(project);
+    zipFile.generateAsync({type: "blob", mimeType: "application/zip"}).then((blob: any) => {
+      const link = document.createElement("a");
+      link.download = zipName;
+      link.href = URL.createObjectURL(blob);
+      link.click();
```
yurydelendik: it does not work on FF -- can you add `link` to the DOM and then remove it

yurydelendik:
- [ ] test on Firefox

src/service.ts
haxxorsid: You mean by doing something like:
`document.body.appendChild(link); `
`link.click();`
`document.body.removeChild(document.body.lastChild);`

Right?

src/service.ts
```diff
@@ -331,6 +332,39 @@ export class Service {
...
+      }
+    }
+    defineZip(project);
+    zipFile.generateAsync({type: "blob", mimeType: "application/zip"}).then((blob: any) => {
+      const link = document.createElement("a");
+      link.download = zipName;
+      link.href = URL.createObjectURL(blob);
+      document.body.appendChild(link);
+      link.click();
+      document.body.removeChild(document.body.lastChild);
```
yurydelendik: can you change it to `document.body.removeChild(link);` and add a comment why we are doing it, including why we are creating a-tag in first place?

haxxorsid: Yeah, I was doing this...

yurydelendik: actually, can you extract entire code and make a function from it?

src/service.ts
```diff
@@ -331,6 +332,42 @@ export class Service {
...
     return uri;
   }
+
+  static downloadProject(project: Project, uri?: string) {
+    const zipFile: JSZip = new JSZip();
+    let zipName: string = "download.zip";
+    if (!isUndefined(uri)) {
+      zipName = `${uri}.zip`;
+    }
+    function defineZip(file: File, folder?: any) {
```
mbebenita: Is there no type available for the `folder` parameter?

src/service.ts
```diff
@@ -331,6 +332,42 @@ export class Service {
...
+        file.mapEachFile((file: File) => defineZip(file, folder));
+      } else {
+        if (!isUndefined(folder)) {
+          folder.file(file.name, file.data);
+        } else {
+          zipFile.file(file.name, file.data);
+        }
+      }
+    }
+    defineZip(project);
```
mbebenita: Does `zipFile` have the same interface as the result of `zipFile.folder(..)`? If so, we can pass `zipFile` here as the `folder` parameter, and remove the special case logic in the `defineZip` function.

src/service.ts
mbebenita: The name `defineZip` is not very clear. This function is adding zip files recursively to the archive. Perhaps we can call it `addZipFile`, or `addZipFileRecursively`, or `zip`, etc.

src/service.ts
```diff
@@ -331,6 +332,42 @@ export class Service {
...
+    zipFile.generateAsync({type: "blob", mimeType: "application/zip"}).then((blob: any) => {
```
mbebenita: Use `await` here.

src/service.ts
haxxorsid: It is casted as `object`. But if defined as an object, it shows up this:
`Property 'file' does not exist on type 'object'` (here property `file` means the .file func used for adding files). So I better thought to leave it as `any`.

src/service.ts
```diff
@@ -331,6 +332,40 @@ export class Service {
...
+          folder = zipFile.folder(file.name);
+          file.mapEachFile((file: File) => addZipFile(file, folder));
+        } else {
+          file.mapEachFile((file: File) => addZipFile(file, zipFile));
+        }
+      } else {
+        folder.file(file.name, file.data);
+      }
+    }
+    addZipFile(project);
```
yurydelendik: remove addZipFile recursion by creating a queue (below is a pseudo code) and make JSZip API comfomant:

```
   var queue = [];
   queue.push({filePrefix: '', file: project});
   while (queue.length > 0) {
      const { filePrefix, file } = queue.pop();
      const fileName = filePrefix + file.name;
      if (file instanceof Directory) {
         file.mapEachFile(f => queue.push({filePrefix: fileName + '/', file: f}));
         if (!(file instance Project)) zipFile.folder(fileName);
         continue;
      }
      zipFile.file(fileName, file.data);
   }
```

src/service.ts
```diff
@@ -331,6 +332,40 @@ export class Service {
...
+    await zipFile.generateAsync({type: "blob", mimeType: "application/zip"}).then((blob: any) => {
```
yurydelendik: Specify `blob` type

src/service.ts
haxxorsid: const queue: Array<object> = [];
queue.push({filePrefix: "", file: project});
while (queue.length > 0) {
    const { filePrefix, file } = queue.pop(); //problem here
    ...
}

queue.pop is returning `ƒ pop() { [native code] }`
But queue becomes properly defined as: [{filePrefix:'', file: project}]  at L2

yurydelendik: > queue.pop is returning ƒ pop() { [native code] }

That's correct, `queue.pop`. What is returned by `queue.pop()` though? Actually, use `queue.shift()` instead.

haxxorsid: Correction: queue.pop() returns that function. Same for queue.shift().

yurydelendik: Really, is it how TS parses `{filePrefix: "", file: project}` ? Okay try `{filePrefix: "", file: project,}`

haxxorsid: Tried that, it didn't worked.

queue[0] returns `{filePrefix: "", file: project}`
queue.pop() does its job, it pops out.

But: const { filePrefix, file } = queue[0] doesn't work also. 
This is strange hmm hmm. There is some problem with data-structure.

yurydelendik: Worked for me when queue type is defined as `var queue: Array<{filePrefix: string; file: File}> = [];`

yurydelendik: Working code

```
    var queue: Array<{filePrefix: string; file: File}> = [];
    project.mapEachFile((f: File) => queue.push({filePrefix: '', file: f}));
    while (queue.length > 0) {
       const { filePrefix, file } = queue.shift();
       const fileName = filePrefix + file.name;
       if (file instanceof Directory) {
          file.mapEachFile((f: File) => queue.push({filePrefix: fileName + '/', file: f}));
          if (!(file instanceof Project)) zipFile.folder(fileName);
          continue;
       }
       zipFile.file(fileName, file.data);
    }
```

src/service.ts
```diff
@@ -331,6 +332,40 @@ export class Service {
...
+    }
+    const queue: Array<{filePrefix: string; file: File}> = [];
+    project.mapEachFile((f: File) => queue.push({filePrefix: "", file: f}));
+    while (queue.length > 0) {
+      const {filePrefix, file} = queue.pop();
+      const fileName = filePrefix + file.name;
+      if (file instanceof Directory) {
+        file.mapEachFile(f => queue.push({filePrefix: fileName + "/", file: f}));
+        if (!(file instanceof Project)) {
+          zipFile.folder(fileName);
```
yurydelendik: use this statement without if, since we have only one Project at the root.

src/service.ts
```diff
@@ -331,6 +332,40 @@ export class Service {
...
+  static async downloadProject(project: Project, uri?: string) {
+    const zipFile: JSZip = new JSZip();
+    let zipName: string = "download.zip";
+    if (!isUndefined(uri)) {
+      zipName = `${uri}.zip`;
+    }
+    const queue: Array<{filePrefix: string; file: File}> = [];
+    project.mapEachFile((f: File) => queue.push({filePrefix: "", file: f}));
+    while (queue.length > 0) {
+      const {filePrefix, file} = queue.pop();
```
yurydelendik: change "pop" to "shift" to be consistent with "queue" algorithm.

src/service.ts
```diff
@@ -331,6 +332,38 @@ export class Service {
...
+      const {filePrefix, file} = queue.shift();
+      const fileName = filePrefix + file.name;
+      if (file instanceof Directory) {
+        file.mapEachFile(f => queue.push({filePrefix: fileName + "/", file: f}));
+        zipFile.folder(fileName);
+        continue;
+      }
+      zipFile.file(fileName, file.data);
+    }
+    await zipFile.generateAsync({type: "blob", mimeType: "application/zip"}).then((blob: object) => {
```
yurydelendik: Change `(blob: object)` to `(blob: Blob)`

yurydelendik:
thank you for the patch

PR #51, opened 2018-02-09
Add file to the targeted dir

haxxorsid:
### Summary of Changes

* Add file to the targeted dir when using New File option.
* Removed redundant if condition from Upload File

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -632,7 +632,7 @@ export class App extends React.Component<AppProps, AppState> {
             this.setState({ newFileDialogDirectory: null });
           }}
           onCreate={(file: File) => {
-            this.project.addFile(file);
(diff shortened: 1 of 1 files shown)


src/components/App.tsx
```diff
@@ -632,7 +632,7 @@ export class App extends React.Component<AppProps, AppState> {
             this.setState({ newFileDialogDirectory: null });
           }}
           onCreate={(file: File) => {
-            this.project.addFile(file);
+            this.state.newFileDialogDirectory.addFile(file);
```
yurydelendik: It's coming from previous code, but you cannot mutate state this way. `this.state.uploadFileDialogDirectory` needs to be refactored or PR that introduced it removed from the code.

PR #53, opened 2018-02-10
TypeScript support / AssemblyScript

dcodeIO:
This PR adds a new "Empty TypeScript Project" option to the initial project selection. Compilation is performed in the build script (no server-side compilation) using the [distribution files](https://github.com/AssemblyScript/assemblyscript/tree/master/dist) required directly from the [AssemblyScript repo](https://github.com/AssemblyScript/assemblyscript) (for now).

Opening this PR as a proof of concept to see what you think?

TODO:

* Pipe `stderr` to the output window (haven't yet figured out how)
* Implement JSON output of parse/compile issues on the AssemblyScript side and display these in the problems window
* Ship a version of assembly.d.ts somehow for proper highlighting and completions


dcodeIO:
cc @MaxGraey Maybe something you are interested in as well?

mbebenita:
Wow, this is great! You are using this in a way that I hadn't expected, and it makes total sense.

I'm curious, shouldn't this be "Empty AssemblyScript Project"?

We can expose a reference to the output window, so you can write to that in the template. To deal with errors, you'll just have to [set the problems property](https://github.com/wasdk/WebAssemblyStudio/blob/master/src/model.ts#L341) on the source file, the UI should pick those up and display them.

Also, we already integrate Binaryen.js for optimization and validation. Is it possible to reuse it in this context?

dcodeIO:
> I'm curious, shouldn't this be "Empty AssemblyScript Project"?

Thought about this as well. The source language is still TypeScript though, but the compiler is AssemblyScript. I'm fine with both :)

> We can expose a reference to the output window, so you can write to that in the template.

That would be great!

> Also, we already integrate Binaryen.js for optimization and validation. Is it possible to reuse it in this context?

As long as it is guaranteed to be up to date with the nightly builds produced by this [binaryen.js](https://github.com/AssemblyScript/binaryen.js) buildbot, yeah. Currently, whenever a new feature is added to the Binaryen-C/JS API, asc immediately makes use of it most of the time.

mbebenita:
@dcodeIO you can now cal `logLn` to write into the Output window. 

```function logLn(message: string, kind?: "" | "info" | "warn" | "error"): void```

Landed in: https://github.com/wasdk/WebAssemblyStudio/commit/9967bd78aa3bfd080e3ae3e081a34a8758163f6d

**Comment by mbebenita** [comment]
> Ship a version of assembly.d.ts somehow for proper highlighting and completions

This is how I get the completions working in the `build.ts` file: https://github.com/wasdk/WebAssemblyStudio/blob/master/src/components/App.tsx#L271

Perhaps we can make the `addExtraLib` available in the `build.ts` file.

**Comment by mbebenita** [comment]
@dcodeIO let me know when you feel the PR is done, and I can merge it.

File: templates/empty_ts/build.ts [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+        logLn(">> Writing file: " + filename);
+        project.newFile(filename.substring(1), "wasm").setData(contents)
+      },
+      listFiles: dirname => { /* TODO */ }
+    }, err => {
+      const output = stderr.toString();
+      if (output.length)
+        logLn("\n" + output);
+      if (err)
+        logLn("ERROR: " + err + "\n");
```
mbebenita: You can pass "error" as the second argument, this will also give you some fancy color highlighting in the output window. No need to pass `\n`, it should be added by default.

templates/templates.js
```diff
@@ -60,4 +60,35 @@
       ["README.md"]
     ]
   }
+}, {
+  "name": "Empty AssemblyScript Project",
+  "description": `# Empty AssemblyScript Project
```
mbebenita: The contents of this field is shown when selecting the template on the main screen. It's a good opportunity to describe what AssemblyScript is all about, maybe add a link back to the project, etc.

dcodeIO:
> Perhaps we can make the addExtraLib available in the build.ts file.

Thanks, `addExtraLib` sounds like the correct approach here. Though, a user would have to run "Build" once to apply it. Is there another way to run a line of code when "Empty AssemblyScript project" is select?

mbebenita:
> Is there another way to run a line of code when "Empty AssemblyScript project" is selected?

Not yet, but we could call a "load" task within `build.ts` when the project loads, and you can call `addExtraLib` there.

templates/empty_ts/build.ts
```diff
@@ -3,17 +3,20 @@ gulp.task("build", async () => require(["asc"], asc => {
...
     "main.ts",
     "--baseDir", "src",
+    "--textFile", "../out/main.wast",
     "--binaryFile", "../out/main.wasm",
+    "--asmjsFile", "../out/main.asm.js",
     "--sourceMap",
     "--measure"
   ];
 
+  project.setStatus("Building Project ...");
```
mbebenita: I don't think it's necessary to set the status here, a call to `setStatus` appears in the `App.tsx`  `build` function.

dcodeIO: Yeah, already removed it :)

dcodeIO:
> Not yet, but we could call a "load" task within build.ts when the project loads, and you can call addExtraLib there.

Sounds good, just added such a task. Should work as soon as the load task is called!

dcodeIO:
I figured that a transparent onload hook might be better here, so the gulp code remains clean and reusable. Essentially there is a new property `onload` on project templates now that takes some code that is being executed when the project is created. Let me know what you think :)

Other than that this should be good to merge now.

src/components/App.tsx
```diff
@@ -619,6 +631,13 @@ export class App extends React.Component<AppProps, AppState> {
               this.logLn("Template doesn't contain a project definition.", "error");
             } else {
               const json = await Service.loadProject(template.project, this.project);
+              if (template.onload) {
```
mbebenita: See note below on handling templates. Also, calling the `onload` logic should also happen for fiddles that are loaded with an Id, not just created from a template.

templates/templates.js
```diff
@@ -60,4 +60,57 @@
...
+      "type": "typescript",
+    }, {
+      "name": "README.md",
+      "type": "markdown",
+    }],
+    "openedFiles": [
+      ["README.md"]
+    ]
+  },
+  "onload": `
```
mbebenita: Templates should not be special in any way, they should more or less behave like fiddles others create with the exception that they are shown in a list when no fiddle is specified. In other words, we can't attach logic to templates that cannot be added to fiddles.

dcodeIO:
Alright, this uses a bit of trickery now to have the setup code a bit separated. Let me know what you think :)

mbebenita:
Thanks. I will try to merge this as a sequence of small independent commits. Follow up work should appear in a new PR.

mbebenita:
Landed in #64

Issue #52, opened 2018-02-10
Keep bundle.js under 500Kb unminified.


mbebenita:
The current size is 1.18 MB, webpack-bundle-size-analyzer prints out:

```
pako: 213.83 KB (18.9%)
wasmparser: 160.01 KB (14.2%)
jszip: 140.69 KB (12.4%)
  core-js: 9.31 KB (6.62%)
  <self>: 131.38 KB (93.4%)
readable-stream: 66.22 KB (5.86%)
buffer: 47.47 KB (4.20%)
cassowary: 47.01 KB (4.16%)
react-modal: 33.15 KB (2.93%)
mousetrap: 32.28 KB (2.85%)
prop-types: 24.88 KB (2.20%)
minimatch: 24.75 KB (2.19%)
util: 16.05 KB (1.42%)
  inherits: 672 B (4.09%)
  <self>: 15.4 KB (95.9%)
events: 8.13 KB (0.719%)
string_decoder: 8.1 KB (0.717%)
setimmediate: 6.32 KB (0.559%)
path-browserify: 6.04 KB (0.534%)
lie: 5.79 KB (0.512%)
process: 5.29 KB (0.468%)
brace-expansion: 4.68 KB (0.414%)
fbjs: 4.26 KB (0.377%)
stream-browserify: 3.54 KB (0.313%)
base64-js: 3.39 KB (0.300%)
core-util-is: 2.95 KB (0.261%)
object-assign: 2.06 KB (0.182%)
ieee754: 2.01 KB (0.177%)
immediate: 1.84 KB (0.163%)
warning: 1.76 KB (0.156%)
util-deprecate: 1.58 KB (0.139%)
safe-buffer: 1.49 KB (0.132%)
timers-browserify: 1.33 KB (0.117%)
balanced-match: 1.15 KB (0.101%)
process-nextick-args: 1.02 KB (0.0898%)
webpack: 1 KB (0.0886%)
exenv: 863 B (0.0745%)
inherits: 672 B (0.0580%)
concat-map: 345 B (0.0298%)
isarray: 132 B (0.0114%)
<self>: 248.61 KB (22.0%)
```

mbebenita:
@haxxorsid jszip is too large to be included in `bundle.js`, especially since it's not used until downloading is requested. Can you move it into a dynamically loaded module? Or work out a simpler way to package files for downloading.

yurydelendik:
> Can you move it into a dynamically loaded module?

[...]

PR #54, opened 2018-02-10
Dynamic import

zurfyx:
Associated Issue: #52

### Result

**Previous**
```
Asset     Size  Chunks                    Chunk Names
bundle.js  1.01 MB       0  [emitted]  [big]  main
bundle.js.map  1.33 MB       0  [emitted]         main
```

**Current**

```
0.bundle.js  534 kB       0  [emitted]  [big] <-- JSZip 
main.bundle.js  522 kB       1  [emitted]  [big]  main
0.bundle.js.map  652 kB       0  [emitted]
main.bundle.js.map  682 kB       1  [emitted]         main
```

### Summary of Changes

**package.json**

-> Removed build-watch

While attempting to dynamically load, having both `webpack-dev-server` and `build-watch` did cause conflicts, such as `Cannot read property 'call' of undefined at __webpack_require__`. `dev-server` will take care of the bundle distribution now for development environment.

**webpack.config.js**

Files will be served by default from `/dist/` folder.

^ Although that works, I suggest switching the whole public path onto dist/, including `index.html` (we can do that through `devServer` webpack config). This way both GHPages and dev-server wouldn't need to server everything, including source code files and configuration. For GHPages, I think it would be as simple as `cd`ing inside the `/dist` folder `before_deploy`.

**download.service.ts**

This is the main reason why of the WIP. I suggest having something like `/services`, and maybe another folder with the dynamically loaded modules, or something like that. Let me know what you think.

**tsconfig.json & tsconfig.test.json**

The previous tsconfig.json set to `target: es6` transpiled `import` to `require`. [By setting it to `esnext`](https://blog.josequinto.com/2017/06/29/dynamic-import-expressions-and-webpack-code-splitting-integration-with-typescript-2-4/#Unexpected-configuration-for-Code-Splitting-with-webpack), webpack is able to identify the `import` calls and generate the chunks.

### Test Plan

WIP

### DIFF
--- a/index.html
+++ b/index.html
@@ -52,6 +52,6 @@ SOFTWARE.
 
         
         <script src="./lib/base64js.min.js"></script>
-        <script src="./dist/bundle.js"></script>
+        <script src="./dist/main.bundle.js"></script>
     </body>
 </html>
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -425,7 +425,8 @@ export class App extends React.Component<AppProps, AppState> {
   }
   async download() {
     this.logLn("Downloading Project ...");
-    await Service.downloadProject(this.project, this.state.fiddle);
+    const downloadService = await import("../utils/download");
+    await downloadService.downloadProject(this.project, this.state.fiddle);
     this.logLn("Project Zip CREATED ");
   }
   /**
--- a/src/service.ts
+++ b/src/service.ts
@@ -25,7 +25,6 @@ import "monaco-editor";
 import { padLeft, padRight, isBranch, toAddress, decodeRestrictedBase64ToBytes } from "./util";
 import { assert } from "./index";
 import getConfig from "./config";
-import * as JSZip from "jszip";
 
 declare interface BinaryenModule {
   optimize(): any;
@@ -318,37 +317,6 @@ export class Service {
       }
     }
     return uri;
-  }
-
-  static async downloadProject(project: Project, uri?: string) {
-    const zipFile: JSZip = new JSZip();
-    let zipName: string = "download.zip";
-    if (!isUndefined(uri)) {
-      zipName = `${uri}.zip`;
-    }
-    const queue: Array<{filePrefix: string; file: File}> = [];
-    project.mapEachFile((f: File) => queue.push({filePrefix: "", file: f}));
-    while (queue.length > 0) {
-      const {filePrefix, file} = queue.shift();
-      const fileName = filePrefix + file.name;
-      if (file instanceof Directory) {
-        file.mapEachFile(f => queue.push({filePrefix: fileName + "/", file: f}));
-        zipFile.folder(fileName);
-        continue;
-      }
-      zipFile.file(fileName, file.data);
-    }
-    await zipFile.generateAsync({type: "blob", mimeType: "application/zip"}).then((blob: Blob) => {
-      // Creating <a> to programmatically click for downloading zip via blob's URL
-      const link = document.createElement("a");
-      link.download = zipName;
-      link.href = URL.createObjectURL(blob);
-      // A fix for making link clickable in Firefox
-      // Explicity adding link to DOM for Firefox
-      document.body.appendChild(link);
-      link.click();
-      document.body.removeChild(link);
-    });
   }
 
   static async exportProjectToGist(project: Project, uri?: string): Promise<string> {
--- a/src/utils/download.ts
+++ b/src/utils/download.ts
@@ -1,0 +1,55 @@
+/* Copyright 2018 Mozilla Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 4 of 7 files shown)


package.json
```diff
@@ -5,8 +5,7 @@
   "main": "index.js",
   "scripts": {
     "build": "webpack --config webpack.config.js",
-    "build-watch": "webpack --watch --config webpack.config.js",
-    "dev-server": "webpack-dev-server --config webpack.config.js",
+    "dev-server": "webpack-dev-server --config webpack.config.js --watch",
```
AnshulMalik: Isn't watch enabled by default?

zurfyx: That's right!

AnshulMalik:
I was just fiddling around, I see that you are trying to refactor a separate download service.
I'd be good if we get a named module like `jszip.bundle.js`

```diff
diff --git a/src/components/App.tsx b/src/components/App.tsx
index 9f62233..3c74b0b 100644
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -425,8 +425,8 @@ export class App extends React.Component<AppProps, AppState> {
   }
   async download() {
     this.logLn("Downloading Project ...");
-    const downloadService = await import("../download.service");
-    await downloadService.downloadProject(this.project, this.state.fiddle);
+    // const downloadService = await import("../download.service");
+    await Service.downloadProject(this.project, this.state.fiddle);
     this.logLn("Project Zip CREATED ");
   }
   /**
diff --git a/src/download.service.ts b/src/download.service.ts
index 65961fa..505f21e 100644
--- a/src/download.service.ts
+++ b/src/download.service.ts
@@ -21,35 +21,3 @@

 import { File, Project, Directory } from "./model";
 import { isUndefined } from "util";
-import * as JSZip from "jszip";
-
-export async function downloadProject(project: Project, uri?: string) {
-  const zipFile: JSZip = new JSZip();
-  let zipName: string = "download.zip";
-  if (!isUndefined(uri)) {
-    zipName = `${uri}.zip`;
-  }
-  const queue: Array<{filePrefix: string; file: File}> = [];
-  project.mapEachFile((f: File) => queue.push({filePrefix: "", file: f}));
-  while (queue.length > 0) {
-    const {filePrefix, file} = queue.shift();
-    const fileName = filePrefix + file.name;
-    if (file instanceof Directory) {
-      file.mapEachFile(f => queue.push({filePrefix: fileName + "/", file: f}));
-      zipFile.folder(fileName);
-      continue;
-    }
-    zipFile.file(fileName, file.data);
-  }
-  await zipFile.generateAsync({type: "blob", mimeType: "application/zip"}).then((blob: Blob) => {
-    // Creating <a> to programmatically click for downloading zip via blob's URL
-    const link = document.createElement("a");
-    link.download = zipName;
-    link.href = URL.createObjectURL(blob);
-    // A fix for making link clickable in Firefox
-    // Explicity adding link to DOM for Firefox
-    document.body.appendChild(link);
-    link.click();
-    document.body.removeChild(link);
-  });
-}
diff --git a/src/service.ts b/src/service.ts
index 42f311f..c2daa35 100644
--- a/src/service.ts
+++ b/src/service.ts
@@ -319,6 +319,41 @@ export class Service {
     return uri;
   }
 
+  static async downloadProject(project: Project, uri?: string) {
+    const JSZip = await import(/* webpackChunkName: "jszip" */ "jszip");
+
+    const zipFile = new JSZip();
+    let zipName: string = "download.zip";
+    if (!isUndefined(uri)) {
+      zipName = `${uri}.zip`;
+    }
+    const queue: Array<{filePrefix: string; file: File}> = [];
+    project.mapEachFile((f: File) => queue.push({filePrefix: "", file: f}));
+    while (queue.length > 0) {
+      const {filePrefix, file} = queue.shift();
+      const fileName = filePrefix + file.name;
+      if (file instanceof Directory) {
+        file.mapEachFile(f => queue.push({filePrefix: fileName + "/", file: f}));
+        zipFile.folder(fileName);
+        continue;
+      }
+      zipFile.file(fileName, file.data);
+    }
+    await zipFile.generateAsync({type: "blob", mimeType: "application/zip"}).then((blob: Blob) => {
+      // Creating <a> to programmatically click for downloading zip via blob's URL
+      const link = document.createElement("a");
+      link.download = zipName;
+      link.href = URL.createObjectURL(blob);
+      // A fix for making link clickable in Firefox
+      // Explicity adding link to DOM for Firefox
+      console.log("Downloading");
+      document.body.appendChild(link);
+      // link.click();
+      document.body.removeChild(link);
+    });
+  }
+
+
   static async exportProjectToGist(project: Project, uri?: string): Promise<string> {
     const files: any = {};
     function serialize(file: File) {
```

You just need to focus at 
```
const JSZip = await import(/* webpackChunkName: "jszip" */ "jszip");
```

zurfyx:
@AnshulMalik Yes, you're right. That does fix the specific problem described in the issue.

However, as the application keeps getting bigger, we're likely to see this same problem with our own code. Generally we would get around this by chunk splitting routes, but given that its all about a single URL, I think it's good to look for alternatives. Or maybe that's just overkill for the current state of the app? :thinking:

AnshulMalik:
Yeah, looks like overkill

mbebenita:
One of the use cases of this application is to be embedded into other sites using IFrames, for example we already have integrated into [http://embed.ly](http://embed.ly) so [http://medium.com](http://medium.com) works just by pasting a URL to the fiddle.

Example: [https://medium.com/@mbebenita/draft-webassembly-studio-22f4eb37597b](https://medium.com/@mbebenita/draft-webassembly-studio-22f4eb37597b)

A post could include several copies of WebAssembly.Studio, so it's very important that we reduce the initial load size to a minimum. Part of the reason I decided to write all UI widgets by hand and not rely on external libraries.

zurfyx:
@mbebenita and do you think that we should also split by external libraries now, or also keep as separate chunks all of the features that aren't really compile, build and run?

AnshulMalik:
Yeah, having libraries split, caching can be more effective. 👍

mbebenita:
@zurfyx which external libraries are you referring to?

zurfyx:
@mbebenita JSZip for example

yurydelendik:
> A post could include several copies of WebAssembly.Studio, so it's very important that we reduce the initial load size to a minimum. 

@mbebenita webpack's code split does just that

package.json
```diff
@@ -5,7 +5,6 @@
   "main": "index.js",
   "scripts": {
     "build": "webpack --config webpack.config.js",
-    "build-watch": "webpack --watch --config webpack.config.js",
```
yurydelendik: unrelated to the PR change?

zurfyx: @yurydelendik It was "incompatible" with the `dev-server`, that's why I got rid of it (mentioned above). But maybe you might want to use it individually?

yurydelendik: What I am asking if it's necessary for this PR? If not, I would like not review this change.

yurydelendik:
Thank you for your patch!

PR #55, opened 2018-02-10
New directory dialog and renamed addFile func

haxxorsid:
### Summary of Changes

* Added new directory dialog
* Renamed addFile function to addFileOrDirectory
* Replaced addFile with addFileOrDirectory everywhere


mbebenita:
@haxxorsid thanks, this looks great! However, `addFile` is a API breaking change because there may be fiddles that depend on it, the `addFile` function appears in `build.ts` files. Also, it's common to think of Directories as Files, so `addFile` is not such a terrible name. (although I admit that addFileOrDirectory is more descriptive). I would revert that change, or at least add back the `addFile` function and have it call `addFileOrDirectory`.

mbebenita:
Thanks!

PR #56, opened 2018-02-10
No edit and delete option for Project

haxxorsid:
### Summary of Changes

* Avoided edit, delete option when target is Project

### Screenshots

* Currently, when the project is tried to be edited App breaks, no need of editing Project at all:
Edit breaks the App
* Currently, when delete option is used on the project it shows an alert but doesn't work, it is better not to make it work:
Deleting whole project


mbebenita:
👍 thanks!

Issue #57, opened 2018-02-10
Double clicking on any tab pins the preview tab.


mbebenita:

haxxorsid:
`this.state.group.open(file, false);`
Hmm.

dcodeIO:
Is this still an issue after https://github.com/wasdk/WebAssemblyStudio/pull/87 ? If so, could you explain a bit how to reproduce (not sure what the "preview" tab is)?

haxxorsid:
![Preview Tab](https://snag.gy/HUw0lh.jpg)
Open some tab(single click) on a file from directorytree, the tab's name will be italic. That italic one will be the preview tab, if I try to single click on some other file from directorytree, that other file will be opened up in the preview tab.

mbebenita:
A "preview tab" is the tab in which new files will be loaded, for instance:

![image](https://user-images.githubusercontent.com/311082/36275262-349584b2-123f-11e8-833f-5b22762dc685.png)

the tab for "main.js" is the preview tab. If you click on another file in directory tree, it will open in this tab unless it's already been opened in a tab. There is only one preview tab for a group. If there is no preview tab, the next file you open will open in a new preview tab. The point of all this is to reduce the number of tabs that are created. 

There are two ways to make the tab no longer preview (in other words, pin it). You can do that by double clicking the tab itself, or double clicking the file in the directory tree. The problem at the moment, is that you can pin the `main.js` file by double clicking on `main.c`.

mbebenita:
Ah, @haxxorsid and I tried to answer at the same time.

dcodeIO:
Thanks to both of you :)

haxxorsid:
Close this.

PR #60, opened 2018-02-10
Double clicking on tab won't pin the preview tab

haxxorsid:
Associated Issue: #57

### Summary of Changes

* Passed nothing(defaultly true to open function of Group)

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -570,7 +570,7 @@ export class App extends React.Component<AppProps, AppState> {
             if (file instanceof Directory) {
               return;
             }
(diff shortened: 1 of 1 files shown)


mbebenita:
Double clicking should pin the tab, but only when clicking on the preview tab. The problem here is that if you double click on other tabs, the preview tab also gets pinned.

haxxorsid:
Oh Oh

haxxorsid:
Double-clicking on other tab is not pinning  tab on my setup. Tried in firefox nightly and chrome.

PR #69, opened 2018-02-12
Reuse filetypeForExtension in AssemblyScript setup code

dcodeIO:
Just a little clean-up for `setup.js` used in AssemblyScript projects. Also correctly uses a `FileType` argument instead of `Language` now.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -36,7 +36,7 @@ import {
   updateFileNameAndDescription,
   deleteFile,
 } from "../actions/AppActions";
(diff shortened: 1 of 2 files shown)


yurydelendik:
Thank you for the patch!

Notes for future:
- We shall not pass `this.state.project.getModel()` as is -- model is not allowed to be mutated in App.tsx
- `filetypeForExtension` needs to be replaced with `getTypeForFilename`, e.g. `Makefile` might not have extension.

Issue #74, opened 2018-02-13
Gist context menu option.


mbebenita:
We can add a gist option in the context menu when right clicking on a file, or folder. This would make it possible to just share individual files.

yurydelendik:
Fixed by #76

PR #76, opened 2018-02-13
Gist context menu for individual files/folders

haxxorsid:
Associated Issue: #74

### Summary of Changes

* Added gist context menu
* Modified gist function(for reusing) to except either file/directory or nothing, if nothing then grab the whole project.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -393,10 +393,10 @@ export class App extends React.Component<AppProps, AppState> {
     }
     this.setState({ fiddle });
   }
-  async gist() {
+  async gist(fileOrDirectory?: File) {
     this.logLn("Exporting Project ...");
-    const projectModel = this.state.project.getModel();
-    const gistURI = await Service.exportProjectToGist(projectModel, this.state.fiddle);
+    const target: File = fileOrDirectory || this.state.project.getModel();
+    const gistURI = await Service.exportToGist(target, this.state.fiddle);
     this.logLn("Project Gist CREATED ");
     if (gistURI) {
       if (this.toastContainer) {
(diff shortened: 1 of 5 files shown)


src/components/App.tsx
```diff
@@ -393,10 +393,15 @@ export class App extends React.Component<AppProps, AppState> {
...
-  async gist() {
+  async gist(fileOrDirectory?: File) {
     this.logLn("Exporting Project ...");
-    const projectModel = this.state.project.getModel();
-    const gistURI = await Service.exportProjectToGist(projectModel, this.state.fiddle);
+    let gistURI: string;
+    if (fileOrDirectory) {
+      gistURI = await Service.exportToGist(fileOrDirectory, this.state.fiddle);
+    } else {
+      const projectModel = this.state.project.getModel();
```
yurydelendik: can you replace the projectModel var with `const target: File = fileOrDirectory || this.state.project.getModel();`?

this means you don't need the if and gistURL change.

src/components/App.tsx
```diff
@@ -393,10 +393,10 @@ export class App extends React.Component<AppProps, AppState> {
     }
     this.setState({ fiddle });
   }
-  async gist() {
+  async gist(fileOrDirectory?: File) {
     this.logLn("Exporting Project ...");
-    const projectModel = this.state.project.getModel();
-    const gistURI = await Service.exportProjectToGist(projectModel, this.state.fiddle);
+    const target: File = fileOrDirectory || this.state.project.getModel();
+    gistURI = await Service.exportToGist(target, this.state.fiddle);
```
yurydelendik: `const ` is missing here

yurydelendik:
Thank you for the patch.

Issue #16, opened 2018-02-04
Favicon is low-res.


mbebenita:

dcodeIO:
Not sure if the PR above solves this issue. If not, what else would be necessary? :)

mbebenita:
I think it solved it! Fixed by #77 

![image](https://user-images.githubusercontent.com/311082/36166167-8e38e134-10a6-11e8-83b3-12aaeb67283f.png)

PR #77, opened 2018-02-13
Updated favicon and colors to be more WebAssembly-y

dcodeIO:
While not really 'high res', the favicon has double the resolution now (see #16). Style now uses webassembly-purple instead of vscode-blue where applicable. Wdyt?


dcodeIO:
img

mbebenita:
Looks great! Thanks!

Issue #34, opened 2018-02-07
Integrate functionality from WasmCodeExplorer


mbebenita:
Currently, the only way to inspect a wasm file is to see its disassembled output. It would be great if we could also let users look at the binary file directly: https://wasdk.github.io/wasmcodeexplorer/

haxxorsid:
I think normally this will slow down the client's browser, maybe crash or lag if he tries to inspect. Later, we will have to make sure all the data gets cleaned up when the client exits the option.

mbebenita:
One temporary fix would be to open WasmCodeExplorer in an pop window and pass the `.wasm` file to it.

yurydelendik:
See also #32

PR #78, opened 2018-02-13
Code-explorer context option for Wasm

haxxorsid:
Associated Issue: #34

### Summary of Changes

* Added context menu for inspecting Binary of Wasms.
* Opens explorer in a new window with the loaded binary of target wasm.

#### Why new window?
In case a huge wasm is passed, it lags up the window or maybe it will lead to a crash. So opening explorer in a new window is a better option (as suggested by mbx).

### DIFF
--- a/src/components/DirectoryTree.tsx
+++ b/src/components/DirectoryTree.tsx
@@ -186,6 +186,9 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
           actions.push(new MonacoUtils.Action("x", "To Firefox x86 Baseline", "octicon-file-binary", true, () => {
             Service.disassembleX86(file, "--wasm-always-baseline");
           }));
+          actions.push(new MonacoUtils.Action("x", "Binary Explorer", "octicon-file-binary", true, () => {
+            Service.openBinaryExplorer(file);
+          }));
         } else if (file.type === FileType.C || file.type === FileType.Cpp) {
           actions.push(new MonacoUtils.Action("x", "Clang-Format", "octicon-quote ruler", true, () => {
             Service.clangFormat(file);
--- a/src/service.ts
+++ b/src/service.ts
@@ -565,6 +565,23 @@ export class Service {
     output.setData(s);
   }
 
+  static openBinaryExplorer(file: File) {
+    const childWindow = window.open(
+      "//wasdk.github.io/wasmcodeexplorer/?api=postmessage",
+      "",
+      "toolbar=no,ocation=no,directories=no,status=no,menubar=no,location=no,scrollbars=yes,resizable=yes,width=1024,height=568"
+    );
+    window.addEventListener("message", function(e: any) {
+      if (e.data.type === "wasmexplorer-ready") {
+        const dataToSend = new Uint8Array((file.data as any).slice(0));
+        e.source.postMessage({
+          type: "wasmexplorer-load",
+          data: dataToSend
+        }, "*", [dataToSend.buffer]);
+      }
+    } , false);
+  }
+
   static async compileMarkdownToHtml(src: string): Promise<string> {
     if (typeof showdown === "undefined") {
       await Service.lazyLoad("lib/showdown.min.js");


src/service.ts
```diff
@@ -552,6 +552,13 @@ export class Service {
...
+    win.document.write(`<!DOCTYPE html><html><head><title>WebAssembly Code Explorer</title><meta charset="utf-8"> <script src="https://cdnjs.cloudflare.com/ajax/libs/systemjs/0.20.12/system.js"></script> <link rel="stylesheet" type="text/css" href="style/code-explorer.css"><link href="https://fonts.googleapis.com/css?family=Roboto|Roboto+Mono" rel="stylesheet"></head><body>`);
+    win.document.write(`<div class="toolbar"><div class="title"> WebAssembly Code Explorer</div> <button id="openFile">Open File</button><div class="hidden"><input type="file" id="browseFile"></div></div><div class="container"><div id="dump" class="dump"><div class="addresses"><div></div></div><div class="rows"><div class="row"></div></div><div class="asciis"><div class="ascii"></div></div></div><div id="text" class="text"></div></div> <script>window["rootArr"]=new Uint8Array([${file.data}]);SystemJS.config({map:{\'wasmparser\':\'https://npmcdn.com/wasmparser/\',},packages:{\'wasmparser\':{format:\'cjs\',main:\'./dist/index.js\',defaultExtension:\'js\',map:{\'stream\':\'data:,exports.Transform=function(){}\',},},}});System.import(\'./lib/code-explorer.js\');</script>`);
```
haxxorsid: This is absolutely barbaric. Any better way to do this?

lib/code-explorer.js
```diff
@@ -0,0 +1,272 @@
...
+        ascii.className = 'ascii';
+        ascii.textContent = str;
+        asciis.appendChild(ascii);
+    }
+    paintCode(dump.querySelectorAll('.o'));
+    disassemble();
+}
+
+initialize();
+openWasm(window["rootArr"]);
```
haxxorsid: I have saved Uint8Array in window["rootArr"] (assigned in HTML). Any other better way to access, assign variable?

src/service.ts
yurydelendik: > This is absolutely barbaric. Any other way to do this?

save it as a static file (e.g. code-explorer.html) and open it as a such?

lib/code-explorer.js
yurydelendik: > Any better way to access file data?

In this context, postMessage and message event will work just fine.

haxxorsid: services.ts:
```
static openBinaryExplorer(file: File) {
    let child_window = null;
    child_window = window.open("./lib/code-explorer.html", "", "toolbar=no,ocation=no,directories=no,status=no,menubar=no,location=no,scrollbars=yes,resizable=yes,width=1024,height=568");
    child_window.postMessage(file.data, "*");
  }
```

code-explorer.js:
```
window.addEventListener('message', function(e) {
	console.log("event happened");
	recievedFileData(e);
} , false);

function recievedFileData(message) {
	console.log(message);
}
```

I am recieving rubbish in message.data:
Screenshot

yurydelendik: it's probably coming from react's dev server. You will need to filter only one you care, e.g. add the type property on the object you are sending.

lib/code-explorer.js
```diff
@@ -0,0 +1,272 @@
+var wasmparser = require('wasmparser');
```
yurydelendik: If you copied code from wasmcodeexplorer, you can put MIT license header here -- I'm the author.

lib/code-explorer.js
haxxorsid: Euhm...what about data is not being received. No message containing file binary is visible in my logs.

yurydelendik: use load events to track when other window is load, or notify the parent with postMessage when it's ready for data

haxxorsid:
@yurydelendik Take a glimpse at latest commit.

lib/code-explorer.js
```diff
@@ -0,0 +1,302 @@
+/* Copyright 2018 Yury Delendik
```
yurydelendik: Can you change that to Mozilla Foundation?

lib/code-explorer.js
```diff
@@ -0,0 +1,302 @@
...
+}
+
+initialize();
+window.onload = function() {
+    window.opener.postMessage({type: "ready"}, "*");
+}
+window.addEventListener("message", function(e) {
+    if (e.data.type === "file") {
+        //console.log checking
+        console.log(e.data.file);
```
yurydelendik: here and below remove console.log

src/components/DirectoryTree.tsx
```diff
@@ -153,6 +153,9 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
           actions.push(new (window as any).Action("x", "Firefox x86 Baseline", "octicon-file-binary", true, () => {
             Service.disassembleX86(file, "--wasm-always-baseline");
           }));
+          actions.push(new (window as any).Action("x", "Binary Explorer", "octicon-file-binary", true, () => {
```
yurydelendik: it's MonacoUtils.Action now

src/service.ts
```diff
@@ -552,6 +552,16 @@ export class Service {
     output.setData(s);
   }
 
+  static openBinaryExplorer(file: File) {
+    const childWindow = window.open("./lib/code-explorer.html", "", "toolbar=no,ocation=no,directories=no,status=no,menubar=no,location=no,scrollbars=yes,resizable=yes,width=1024,height=568");
```
yurydelendik: Not sure if we want to open a new window. I think, similar to #67, we want to open wasm as some preview. @mbebenita may have different idea for wasm file?

mbebenita: Ah, opening as a tab would probably be better, but i think we need to first land the MD preview work, and make this fit into that. I think we should land this and live with a popup for a while, then integrate into a tab later.

yurydelendik:
I cannot make it work locally:

screen shot 2018-02-15 at 4 11 10 pm

(Also, we probably don't need "Open File..." in this case)

(Crazy idea: extend https://github.com/wasdk/wasmcodeexplorer with postMessage API you created here, and use only DirectoryTree.tsx/service.ts code to interact with other project HTML page)

haxxorsid:
Won't that make up the same thing?

yurydelendik:
> For the crazy idea: Won't that make up the same thing? 

In this case maintenance of the API falls on different project. But anyway, if it's a short fix, let's do it first.

haxxorsid:
Cool.
_I will update PR in later half of next week, I'm busy this week._

yurydelendik:
I add https://github.com/wasdk/wasmcodeexplorer/commit/512113bfbe854fe48f4321b60f382bf9461231f4 to the explorer and fixed how arraybuffer is passed to opened window.

Thank you for looking into the issue. Fixes #34.

haxxorsid:
@yurydelendik thanks alot.

PR #79, opened 2018-02-13
Move tab groups logic to store and actions

AnshulMalik:
App.tsx currently does a lot of things that it should not, moving tabs logic out of the component will help in better understanding of the code.

### Summary of Changes

* Change groups to tabGroups
* Move tabs out of App.tsx

### Test Plan

- [x] Open file from left file tree(Single and Double click)
- [x] Split pane
- [x] Close file in either pane

### DIFF
--- a/src/actions/AppActions.ts
+++ b/src/actions/AppActions.ts
@@ -21,6 +21,11 @@
 
 import dispatcher from "../dispatcher";
 import { File, Directory, Project } from "../model";
+import { App } from "../components/App";
+import { ProjectTemplate } from "../components/NewProjectDialog";
+import appStore from "../stores/AppStore";
+import { Service } from "../service";
+import Group from "../utils/group";
 
 export enum AppActionType {
   ADD_FILE_TO = "ADD_FILE_TO",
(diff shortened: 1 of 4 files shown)


src/actions/AppActions.ts
```diff
@@ -108,3 +121,76 @@ export function logLn(message: string, kind: "" | "info" | "warn" | "error" = ""
...
+    file
+  } as CloseFileAction);
+}
+
+export interface OpenProjectFilesAction extends AppAction {
+  type: AppActionType.OPEN_PROJECT_FILES;
+  json: ProjectTemplate;
+}
+
+export function openProjectFiles(json: Object) {
```
yurydelendik: make this function async and move:

```
const projectModel = new Project();
const json = await Service.loadProject(project, projectModel);
```
and dispatch `projectModel` and `json.openedFiles`. Notice that ideally we need to create a new project. Similar to load from fiddle (can we use LOAD_PROJECT action for that?)

src/actions/AppActions.ts
```diff
@@ -108,3 +121,76 @@ export function logLn(message: string, kind: "" | "info" | "warn" | "error" = ""
...
+  dispatcher.dispatch({
+    type: AppActionType.OPEN_PROJECT_FILES,
+    json
+  } as OpenProjectFilesAction);
+}
+
+export async function saveProject(fiddle: string) {
+  this.logLn("Saving Project ...");
```
yurydelendik: use logLn

yurydelendik:
Good to go after commit squashing.

yurydelendik:
Thank you for the patch.

mbebenita:
👍

PR #83, opened 2018-02-14
Tidy up directory tree context menu a bit

dcodeIO:
ctx

* Makes sure all icons display with the same width
* Orders by Directory options, Common file options, Gist if applicable, [ruler] File-specific options
* Fixes the first item showing as dark-blue (came from some focus rule in monaco)
* Can now download all kinds of files, not just .wasm
* Strips some not immediately useful detail texts

### DIFF
--- a/src/components/DirectoryTree.tsx
+++ b/src/components/DirectoryTree.tsx
@@ -20,7 +20,7 @@
  */
 
 import * as React from "react";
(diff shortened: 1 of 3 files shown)


mbebenita:
@dcodeIO very nice, thanks! I've been looking for a way to disable that dark-blue focus rule for a while now.

PR #84, opened 2018-02-14
Add favicon source psd

dcodeIO:
This is the source file of the current 32x32 favicon. Not sure if the svg directory is the correct place, but adding in case it might be useful. Note that this isn't 1:1 the WebAssembly logo, but is slightly modified to match pixels so it appears sharper.


mbebenita:
Thanks!

PR #87, opened 2018-02-14
Slightly decrease double click time in directory tree

dcodeIO:
Prevents accidentally opening a lot of files. 500ms is the default on Windows, for example.

### DIFF
--- a/src/components/DirectoryTree.tsx
+++ b/src/components/DirectoryTree.tsx
@@ -307,7 +307,7 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
     if (file instanceof Directory) {
       return;
     }
-    if (Date.now() - this.lastClickedTime < 1000 && this.props.onDoubleClickFile) {
+    if (Date.now() - this.lastClickedTime < 500 && this.props.onDoubleClickFile) {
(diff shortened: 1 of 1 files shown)


dcodeIO:
Oh, wait, it also "double clicks" when clicking two different items. Let me fix that as well.

mbebenita:
👍

src/components/DirectoryTree.tsx
```diff
@@ -303,16 +303,18 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
     });
   }
   lastClickedTime = Date.now();
+  lastClickedFile: File | null = null;
   onClickFile(file: File) {
     if (file instanceof Directory) {
       return;
     }
-    if (Date.now() - this.lastClickedTime < 1000 && this.props.onDoubleClickFile) {
+    if (Date.now() - this.lastClickedTime < 500 && this.lastClickedFile === file && this.props.onDoubleClickFile) {
```
yurydelendik: can move extract a constant, e.g. `const ThrottleClickTimeout = 500;` and place it at the line above? (it will be instead of comment)

PR #89, opened 2018-02-14
Fix tabs background being cut off

dcodeIO:
tabs


yurydelendik:
Thank you for the patch.

PR #93, opened 2018-02-15
Add a few more keywords to 'wast' language

dcodeIO:


yurydelendik:
Thank you for the patch.

PR #94, opened 2018-02-15
Refactor analytics

dcodeIO:
Refactor analytics to its own file and run it only if DNT isn't explicitly requested. Though the right thing to do, this doesn't solve any [technical issues](https://www.w3.org/TR/tracking-dnt/) on its own.

### DIFF
--- a/lib/ga.js
+++ b/lib/ga.js
@@ -1,0 +1,33 @@
+/* Copyright 2018 Mozilla Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


lib/ga.js
```diff
@@ -0,0 +1,12 @@
+if (navigator.doNotTrack != "1") (function() {
+  var script = document.createElement("script");
+  document.head.appendChild(script);
+  script.onload = function() {
+    window.dataLayer = window.dataLayer || [];
+    function gtag(){dataLayer.push(arguments);}
```
yurydelendik: `gtag` needs to exposed on the global/window object. Use `window.gtag = ...`

lib/ga.js
```diff
@@ -0,0 +1,12 @@
+if (navigator.doNotTrack != "1") (function() {
```
yurydelendik: Add a license header.

dcodeIO:
> Please also squash commits.

GitHub also has a Squash-merge ability when selecting the merge action from the merge-button drop down. This results in those fancy "dcodeIO committed with yurydeleendik" commits. Can still squash if that's not an option, just let me know :)

yurydelendik:
> selecting the squash action from the merge-button drop down

this functionality produces commit messages such as "Remove left-over alert" or "Address review comments" which do not make any sense. I prefer to have proper commit comment.

yurydelendik:
Thank you for the patch.

dcodeIO:
> this functionality produces commit messages such as "Remove left-over alert" or "Address review comments" which do not make any sense. I prefer to have proper commit comment.

That's correct, but it still can be edited before finally comitting it. I usually find this useful when merging contributions on my projects because it's just removing a line or two for me, while it is dealing with command line git for the contributor.

yurydelendik:
> while it is dealing with command line git for the contributor

I understand you are not a novice git user :) But for novice git users, I would recommend to use `git commit --amend` for touch up commits (there is also an option in `git gui`). If I see a contributor struggling or ask for it, I'll merge it with squash option.

dcodeIO:
I actually consider myself a novice Git user :) and I must admit that I've never used `--amend`, thanks for the hint.

mbebenita:
I've been stuck at the novice Git user level for 7 years. I usually just send @yurydelendik some garbled up message about what I want to do, and he sends back a command.

PR #96, opened 2018-02-15
Make sure directory tree is properly resized

### LINKED ISSUE #92: DirectoryTree is not changing on resize
STR:
1. open not-tall  (~200px in height) browser window
2. navigate to https://http://webassembly.studio/
3. increase height on the window
4. create a project from a template
5. scroll directory tree up and down

Notice that the tree is not extended all the way to the bottom.

It's expected that tree on the left takes all available space.

dcodeIO:
fixes #92

### DIFF
--- a/src/components/DirectoryTree.tsx
+++ b/src/components/DirectoryTree.tsx
@@ -317,15 +317,8 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
     this.lastClickedFile = file;
 
   }
-  timeout = 0;
-  layout = () => {
-    if (this.timeout) {
-      window.clearTimeout(this.timeout);
-    }
-    this.timeout = window.setTimeout(() => {
-      this.timeout = 0;
-      this.tree.layout();
-    }, 10);
+  onLayout = () => {
+    this.tree.layout();
   }
   componentDidMount() {
     this.ensureTree();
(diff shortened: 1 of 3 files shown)


dcodeIO:
Seems CI didn't catch this one. Local result:

```
Test Suites: 6 passed, 6 total
Tests:       30 passed, 30 total
Snapshots:   0 total
Time:        4.618s
Ran all test suites.

> tslint -p .
```

src/components/DirectoryTree.tsx
```diff
@@ -317,6 +317,16 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
     this.lastClickedFile = file;
 
   }
+  timeout = 0;
+  layout = () => {
```
yurydelendik: can you rename `layout` to something like `onLayout`, `onLayoutDOMEvent`, etc. to avoid confusion with util's `layout` which causes this event?

src/components/DirectoryTree.tsx
```diff
@@ -317,6 +317,16 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
...
+    if (this.timeout) {
+      window.clearTimeout(this.timeout);
+    }
+    this.timeout = window.setTimeout(() => {
+      this.timeout = 0;
+      this.tree.layout();
+    }, 10);
```
yurydelendik: add a constant for magic number 10, e.g. `throttleTimeout`, before `this.timeout =` line.

also, `this.timeout` change to `this.layoutTreeThrottleTimer` provide more "documentation" to the code.

src/components/DirectoryTree.tsx
dcodeIO: Sure. I just copied this code from `components/editor/Editor.tsx`, should I update this one as well?

yurydelendik: No preference. Sure, you can add change in Editor.tsx to this PR.

yurydelendik:
Merged via a69a115 -- fixed 'var' -> 'let'

Thank you for the patch.

PR #103, opened 2018-02-16
If statement for excluding  "out" from exportToGist not required

haxxorsid:
This is not required anymore because: [#97](https://github.com/wasdk/WebAssemblyStudio/pull/97)
### Summary of Changes

* Removed _if_ statement for checking if directory is _out_

#### Note:
I think, in case some user creates a new file inside "_out_" dir manually, then this will export it to gist.


mbebenita:
Ah, thanks!

PR #104, opened 2018-02-16
Added null file data case in loadProject

haxxorsid:
### Summary of Changes
* Set file to empty string, if data is null

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -387,6 +387,8 @@ export class Service {
       file.description = json.description;
       if (json.data) {
         file.setData(json.data);
+      } else if (json.data === null) {
+        file.setData("");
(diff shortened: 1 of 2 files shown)


src/service.ts
```diff
@@ -387,6 +387,8 @@ export class Service {
       file.description = json.description;
       if (json.data) {
         file.setData(json.data);
+      } else if (json.data === null) {
+        file.setData("");
```
mbebenita: It looks like you are trying to ensure that file data is either a valid string, or an array buffer, which is a great idea. To make sure this is always the case please add an assert in the `setData` function. This way all code can safely assume that `data` is not `null` or `undefined`.

Issue #82, opened 2018-02-13
Use .wat instead of .wast?


dcodeIO:
Probably relevant for WebAssembly Studio as well, see: https://github.com/dcodeIO/long.js/issues/53

mbebenita:
I guess we need to change it to stay consistent.

nileshgulia1:
@mbebenita Do we just have to rename `wasm.wast` to `wasm.wat`?

dcodeIO:
There are a couple of mentions of "wast" or "Wast" in file descriptions, comments, the respective language file and internally used functions as well. Should probably rename [these](https://github.com/wasdk/WebAssemblyStudio/search?utf8=%E2%9C%93&q=wast&type=) as well.

haxxorsid:
Close this.

PR #109, opened 2018-02-18
changed wast to wat

nileshgulia1:
Associated Issue: #82 

### Summary of Changes

Changed all wast occurrences to wat and also renamed some files.


mbebenita:
Looks good. One possible concern here is what to do about other projects that may have been saved and shared that use the `.wast` extension.

nileshgulia1:
may be they can be changed too?

PR #110, opened 2018-02-20
Support views

AnshulMalik:
Associated Issue: #17

This is a subset of the main PR of supporting markdown.
This presents another abstraction layer above file, to handle the display related activities and this was also needed so that we can edit a file in two editor instances without having to copy it, and in previewing the file as well.

### Summary of Changes

* Migrate from `file` to `view`
* `EDIT` : Improved the split group logic, the new editor pane now comes right next to current pane cloning the currently active file, previously, it just took the last pane and duplicated it.

### DIFF
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -21,7 +21,8 @@
 
 import * as React from "react";
 import { Project, File, Directory, FileType, languageForFileType } from "../../model";
-import { View, EditorPane } from "./EditorPane";
+import { EditorPane } from "./EditorPane";
+import { View } from "./View";
 import { build, run } from "../../actions/AppActions";
 import "monaco-editor";
 
--- a/src/stores/AppStore.ts
+++ b/src/stores/AppStore.ts
@@ -32,14 +32,17 @@ import {
   LoadProjectAction,
   LogLnAction,
   OpenProjectFilesAction,
-  OpenFileAction,
-  CloseFileAction,
   FocusTabGroupAction,
   SetStatusAction,
-  SandboxRunAction
+  SandboxRunAction,
+  OpenViewAction,
+  CloseViewAction,
+  OpenFileAction
 } from "../actions/AppActions";
 import Group from "../utils/group";
+import { assert } from "../util";
 import { ProjectTemplate } from "../components/NewProjectDialog";
+import { ViewMode, View } from "../components/editor/View";
 
 export class AppStore {
   private project: Project;
@@ -66,7 +69,7 @@ export class AppStore {
 
   private initStore() {
     this.project = new Project();
-    this.activeTabGroup = new Group(null, null, []);
+    this.activeTabGroup = new Group(null, []);
     this.tabGroups = [this.activeTabGroup];
     this.bindProject();
     this.output = new File("output", FileType.Log);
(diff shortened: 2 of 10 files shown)


src/utils/group.ts
```diff
@@ -20,51 +20,70 @@
...
       this.preview = null;
     }
   }
-  close(file: File) {
-    const i = this.files.indexOf(file);
-    assert(i >= 0);
-    if (file === this.preview) {
-      this.preview = null;
+  openFile(file: File, preview = true, mode: ViewMode = ViewMode.EDITOR) {
+    const files = this.views.map(view => view.file);
```
yurydelendik: use `this.views.findIndex(view => view.file === file)` to find the `index`

src/utils/group.ts
```diff
@@ -20,51 +20,70 @@
...
+import { View, ViewMode } from "../components/editor/View";
 import { File } from "../model";
 
 export default class Group {
-  file: File;
-  files: File[];
-  preview: File;
-  constructor(file: File, preview: File, files: File[]) {
-    this.file = file;
+  view: View;
```
yurydelendik: rename `view` to `currentView`

src/utils/group.ts
```diff
@@ -20,51 +20,70 @@
...
+  views: View[];
+  preview: View;
+
+  constructor(view: View, views: View[] = [], preview?: View) {
```
yurydelendik: `preview` argument is not used -- we can remove it

src/utils/group.ts
```diff
@@ -20,51 +20,70 @@
...
 
 export default class Group {
-  file: File;
-  files: File[];
-  preview: File;
-  constructor(file: File, preview: File, files: File[]) {
-    this.file = file;
+  view: View;
+  views: View[];
+  preview: View;
```
yurydelendik: let's make it boolean `isPreview` -- I don't see the case when `this.view !== this.preview`

src/stores/AppStore.ts
```diff
@@ -199,10 +208,11 @@ export class AppStore {
 
   private openProjectFiles(openedFiles: [string[]]) {
     const groups = openedFiles.map((paths: string[]) => {
-      const files = paths.map(file => {
+      const views = paths.map(file => {
         return this.getFileByName(file).getModel();
-      });
-      return new Group(files[0], null, files);
+      }).map((file, i) => new View({ file }));
```
yurydelendik: combine two `map`s into one

src/stores/AppStore.ts
```diff
@@ -199,10 +208,11 @@ export class AppStore {
...
+
+      return new Group(views[0], views);
```
yurydelendik: can you add `assert(views.length > 0)` before this statement?

src/stores/AppStore.ts
```diff
@@ -170,27 +172,34 @@ export class AppStore {
   private splitGroup() {
     const groups = this.tabGroups;
     const lastGroup = groups[groups.length - 1];
-    if (lastGroup.files.length === 0) {
+    if (lastGroup.views.length === 0) {
       return;
     }
-    const group = new Group(lastGroup.file, null, [lastGroup.file]);
+    const group = new Group(lastGroup.view, [lastGroup.view]);
```
yurydelendik: do we clone group here? can we add the `clone()` method, or comment here?

src/utils/group.ts
AnshulMalik: `preview` references to the view which is currently open in preview mode. So it should just be a View.

src/stores/AppStore.ts
AnshulMalik: We are just taking the last file from the group, it's not a full clone, so `clone` would be misleading?

src/utils/group.ts
AnshulMalik: @mbebenita Right now, the only location we call openFile is the files tree, and it's in editor mode. So I left it as such in the hope I don't get too far changing everything as I usually do

src/components/editor/Editor.tsx
```diff
@@ -21,7 +21,7 @@
 
 import * as React from "react";
 import { Project, File, Directory, FileType, languageForFileType } from "../../model";
-import { View, EditorPane } from "./EditorPane";
+import { EditorPane, View } from ".";
```
yurydelendik: Circular reference between, which might cause a confusion about location of the EditorPane and View. Can you use `import { EditorPane } from "./EditorPane";` and `import { View } from "./View";` instead?

src/utils/group.ts
```diff
@@ -20,51 +20,67 @@
...
-      // Switch to file if it's already open.
-      this.file = file;
+      // Switch to the view if it's already open.
+      this.currentView = view;
       if (!shouldPreview) {
         this.preview = null;
       }
-      return;
+
+      return ;
```
yurydelendik: nit: can we remove line break, I don't think this change is needed. (it rather belong after the `if`)

src/utils/group.ts
```diff
@@ -20,51 +20,67 @@
...
+
+    this.views.splice(i, 1);
+    if (this.currentView.onClose) {
+      this.currentView.onClose();
     }
-    this.files.splice(i, 1);
-    this.file = this.files.length ? this.files[Math.min(this.files.length - 1, i)] : null;
+
+    const numViews = this.views.length;
+
```
yurydelendik: nit: remove line here -- it's easier to read since numViews are "attached" to the next statement.

src/utils/group.ts
yurydelendik: but it's essentially the same is the `currentView` it just specifies that `currentView` is in preview mode, is it correct? If yes, I think having a comment about that or name change is nice to have.

src/stores/AppStore.ts
yurydelendik: Can we add "Take the last file from the group" a comment above?

src/utils/group.ts
AnshulMalik: Nope, `currentView` means the view currently being displayed in the pane, and `preview` means the view which is in being previewed (The tab which is shown in italics)

yurydelendik: I'm looking at the code and I see that every time the `currentView` is assigned, the `preview` is assigned to the same exact value or null. Is it correct? If it is correct invariant, we shall enforce it.

Btw, the `close()` is missing such assignment, and that probably a bug since this.preview might stay assigned to the closed view.

AnshulMalik: Yes, preview and currentView are same when we open new files, but in case of switching between the views, you can see that these two are in fact different, in that case only `currentView` changes.

And yes, `close()` is missing this, I'll update

src/utils/group.ts
```diff
@@ -20,51 +20,69 @@
...
     if (index >= 0) {
-      // Switch to file if it's already open.
-      this.file = file;
+      // Switch to the view if it's already open.
+      this.currentView = view;
       if (!shouldPreview) {
         this.preview = null;
       }
-      return;
+      return ;
```
yurydelendik: nit: unneeded change

src/utils/group.ts
```diff
@@ -20,51 +20,69 @@
...
+
+    this.views.splice(i, 1);
+    if (this.currentView.onClose) {
+      this.currentView.onClose();
+    }
+
+    const numViews = this.views.length;
+    this.currentView = numViews ? this.views[Math.min(numViews - 1, i)] : null;
+
+    if(view === this.preview) {
```
yurydelendik: nit: space between `if` and `(` is missing

yurydelendik:
Thank you for the patch.

Issue #119, opened 2018-02-21
Make build.ts a valid gulpfile


yurydelendik:
The example "build.ts" in not executable by gulp. We need the following header:

```
import * as gulp from "gulp";
import { Service, project } from "@wasm/studio-utils";
```

It will be nice to have default package.json included too: with `gulp`, `ts-node` (and `@wasm/studio-utils`) dependencies. Having `gulp --gulpfile ./built.ts` as "build" npm task might be useful as well.

yurydelendik:
Fixed by 5fc2163 and #121

PR #121, opened 2018-02-21
Evolve build.ts into gulpfile.js

dcodeIO:
Associated Issue: #119

### Summary of Changes

* Refactors the "run function in context" functionality to a handy `contextify` utility function that accepts a custom thisArg and context plus custom modules for injection into the `require` chain
* Injects the Gulpy instance as the `gulp` module
* Injects the (proposed?) WebAssembly Studio utilities as the `@wasm/studio-utils` module
* Updates the AssemblyScript template to demonstrate how/that it works
* Retains backwards compatibility as shown by C/Rust/Wat projects (I can update these too with your approval)

### Test Plan

* Create a new AssemblyScript project (uses the new `gulpfile.js`), build and run.
* Create a C/Rust/Wat project (uses the previous `build.ts`), build and run.


dcodeIO:
Btw, I also have the `webassembly` module on npm that we could use as a namespace for npm packages such as `@webassembly/studio-utils`. If you like.

src/actions/AppActions.ts
```diff
@@ -235,18 +236,32 @@ export interface SandboxRunAction extends AppAction {
...
+    }, {
+      // modules
+      "gulp": gulp,
+      "@wasm/studio-utils": {
+        Service,
+        project,
+        logLn,
+        filetypeForExtension,
+        setStatus,
+        clearStatus
```
dcodeIO: Some of these might not be in the corresponding package, but that's what I use as an example.

yurydelendik: Personally, I would like to avoid that custom templates can control UI. Can we remove setStatus/clearStatus and their usage from setup.js?

dcodeIO:
I am using setStatus in the AssemblyScript setup code because it is asynchronous. If a user clicks "Build" too early, this results in a synchronous require error, hence having it there at least provides the information necessary when it's safe to click "build".

Alternative, if there'd a way to enable/disable the "build" button instead, that'd be sufficient as well.

yurydelendik:
> there'd be a way to enable/disable the "build" ... a way to notify the IDE of "some stuff is still loading" 

Can we move the UI discussions and changes to the separate issue/PR? I'll be more comfortable landing gulpfile change without it?

dcodeIO:
Sure

yurydelendik:
Thank you for the patch

PR #122, opened 2018-02-21
Update AssemblyScript project's package.json

dcodeIO:
The only *offline* dependencies ultimately required here are gulp and asc. *Online*, it's pretty much using `project:load` to polyfill the compiler when running in WebAssembly Studio, but this one isn't called in an *offline* environment.


yurydelendik:
Thank you.

dcodeIO:
Thanks :)

PR #123, opened 2018-02-21
Restructure AssemblyScript project for offline use

dcodeIO:
This PR puts the TypeScript sources that the AssemblyScript compiler consumes into their own directory and includes a compatible `tsconfig.json`. Doing so lets an *offline* IDE, for example VSCode, pick up the necessary configuration settings to provide proper hints and highlighting.


mbebenita:
Thanks

Issue #114, opened 2018-02-20
Find a better .wat file icon


mbebenita:
![image](https://user-images.githubusercontent.com/311082/36455469-03679a50-1655-11e8-983f-0748b67b8492.png)

yurydelendik:
Fixed by #126

Issue #65, opened 2018-02-12
Clean up icon handling.


mbebenita:
File icons in the File Tree are from `seti.woff` and defined in `theme.css`. Icons in tabs are loaded from the `svg/` directory, other icons are defined in `Icons.tsx`, and others are defined in `octicons` so it's a bit of a mess.

yurydelendik:
The issues does not specify expected behavior. Looks #126 addressed the main concern though. Closing as fixed.

PR #126, opened 2018-02-22
Unify icons

dcodeIO:
Associated Issues: #65, #114

### Summary of Changes

* Adds a custom seti-extra font containing additional WebAssembly icons
* Updated utilities to return seti icon names
* Changes tabs and various dialogs to use seti-icons
* Picks the correct file icons in the upload file modal.

seti


src/components/NewFileDialog.tsx
```diff
@@ -78,7 +78,7 @@ export class NewFileDialog extends React.Component<{
     return <ReactModal
       isOpen={this.props.isOpen}
       contentLabel="Create New File"
-      className="modal"
+      className="modal show-file-icons"
```
dcodeIO: This is necessary because modals are outer elements, not wrapped by the container with "show-file-icons".

src/components/editor/Tabs.tsx
```diff
@@ -150,13 +150,7 @@ export class Tab extends PureComponent<TabProps, {}> {
...
         onDoubleClick={this.onDoubleClick}
       >
-        {icon && <div
-          className="icon"
-          style={{
-            backgroundImage: `url(svg/${icon}.svg)`
-          }}
-        />
-        }
+        <div className={"monaco-icon-label file-icon " + icon} />
```
dcodeIO: It's ok not to check because the font will resort to the default icon in this case.

style/global.css
```diff
@@ -297,9 +297,12 @@ a {
...
 
-.tab>.icon {
-  width: 16px;
-  height: 16px;
+.tab>.file-icon {
+  vertical-align: middle;
+  text-align: center;
+  font-size: 90%;
+  min-width: 20px;
+  max-height: 20px;
```
dcodeIO: Tab icons require a bit of fixing to be aligned properly. Might be that not all rules are ultimately necessary, but that's where it started to look good.

style/global.css
```diff
@@ -283,7 +283,7 @@ a {
 }
 .tab {
   display: flex;
-  width: 120px;
+  max-width: 150px;
```
dcodeIO: Just enough to handle the "Release Notes" tab, and saves a bit of space where tabs aren't using the full width.

mbebenita:
👍 this is great!

yurydelendik:
Thank you for the patch.

PR #127, opened 2018-02-22
Merge preview

mbebenita:
Lands Markdown preview, with some followup refactoring work. Each markdown View gets a toggle button that lets it switch between Edit / Preview mode.

image

image

### DIFF
--- a/src/components/editor/View.tsx
+++ b/src/components/editor/View.tsx
@@ -35,4 +35,7 @@ export class View {
     this.file = file;
     this.type = type;
   }
+  clone(): View {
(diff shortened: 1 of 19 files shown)


src/components/editor/View.tsx
```diff
@@ -35,4 +35,7 @@ export class View {
     this.file = file;
     this.type = type;
   }
+  clone(): View {
+    return new View(this.file, this.type);
```
AnshulMalik: `clone` generally means making `copy` of an object. But here, the new view is using the reference of the file from previous view ?

PR #133, opened 2018-02-23
Generate production build on npm run build

AnshulMalik:
Associated Issue: #52 

### Summary of Changes

* `npm run build` will now generate a production build
* The size of `main.bundle.js` is checked in the tests

### DIFF
--- a/bin/check-bundle-size.js
+++ b/bin/check-bundle-size.js
@@ -1,0 +1,35 @@
+#!/usr/bin/env node
+
+/*
+  This file was taken from https://github.com/devtools-html/debugger.html/blob/master/bin/ci/check-file-sizes.js
+*/
+
+const path = require("path");
+const fs = require("fs");
+
+const fileSizes = {
+  "main.bundle.js": 300000
+};
+
+const distPath = "./dist";
+
+function checkFileSizes() {
+  let success = true;
+
+  Object.keys(fileSizes).forEach(key => {
+    const fullDistPath = path.join(process.cwd(), distPath);
+    const { size } = fs.statSync(path.join(fullDistPath, key));
+
+    if (size > fileSizes[key]) {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -66,6 +67,7 @@
     "@types/react": "^16.0.28",
     "@types/react-dom": "^16.0.3",
     "awesome-typescript-loader": "^3.4.1",
+    "chalk": "^2.3.0",
```
yurydelendik: let's remove this dependency

bin/check-bundle-size.js
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env node
+
```
yurydelendik: We need some base license test here, or where it came from.

mbebenita:
@AnshulMalik how do I generate a debug build? For development, I use `npm run build-watch`, will this produce a debug build?

AnshulMalik:
Yep @mbebenita , `build-watch` will produce the debug build.

tsconfig.json
```diff
@@ -6,7 +6,7 @@
       "skipLibCheck": true,
       "module": "esnext",
       "moduleResolution": "node",
-      "target": "esnext",
+      "target": "es5",
```
yurydelendik: If we can, let's keep it `es2017`. If uglify-js does not play well, can we use uglify-es or some other minifier? If it's not possible, we probably need to keep code non-production and fix limit to 600k.

bin/check-bundle-size.js
AnshulMalik: I borrowed this file file from `debugger.html` so what should I write as license?

yurydelendik: you can just write where you borrowed the file, since we are not clear about its license.

yurydelendik:
thank you for the patch

Issue #137, opened 2018-02-23
Infinite loop when dragging and dropping a folder over itself.


mbebenita:
STR:

1. Create an Empty C Project.
2. Drag the `src` folder and drop it over itself.
3. Wait forever.

PR #138, opened 2018-02-24
Fix infinite loop caused by dragging and dropping a folder over itself.

ollelauribostrom:
Associated Issue: #137

### Summary of Changes
* Adding if statement in onMoveFile (src/components/App.tsx) that returns early if the file being moved is actually a folder being moved into itself.

### Test Plan
- [x] Create an Empty C Project.
- [x] Drag the src folder and drop it over itself.
- [x] Infinite loop no longer fires

### DIFF
--- a/src/model.ts
+++ b/src/model.ts
@@ -468,6 +468,16 @@ export class File {
   toString() {
     return "File [" + this.name + "]";
   }
+  isDescendantOf(element: File): boolean {
+    let parent = this.parent;
+    while (parent) {
+      if (parent === element) {
+        return true;
+      }
+      parent = parent.parent;
+    }
+    return false;
+  }
 }
 
 export class Directory extends File {
--- a/tests/unit/descendant.spec.ts
+++ b/tests/unit/descendant.spec.ts
@@ -1,0 +1,30 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+import { Directory } from "../../src/model";
+
+function getDirectoryStructure() {
+  const a = new Directory("test");
+  const b = a.newDirectory("b");
+  const c = b.newDirectory("c");
+  return { a, b, c };
+}
+
+describe("File.isDescendantOf tests", () => {
+  it("should return true if file is direct child of given element", () => {
+    const { a, b } = getDirectoryStructure();
+    expect(b.isDescendantOf(a)).toBe(true);
+  });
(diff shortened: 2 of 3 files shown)


mbebenita:
I'd go one step further and add a helper function in the File class, something like `isDescendantOf(parent)`.

mbebenita:
Moreover, it would be better to not allow the file to be dropped on itself in the first place by returning `false` for the `accept` property, see:

https://github.com/wasdk/WebAssemblyStudio/blob/master/src/components/DirectoryTree.tsx#L259

ollelauribostrom:
I made a new commit considering the requested changes / your comments. 

**Changes**:
- Validation is now made in onDragOver (src/components/DirectoryTree.tsx) returning false for the accept property if folder is to be dropped on itself or any child/descendant.
- Implemented a helper function in the File class (src/model.ts) called isDescendantOf that checks if a file is a descendant of a given element

src/model.ts
```diff
@@ -457,6 +457,19 @@ export class File {
   toString() {
     return "File [" + this.name + "]";
   }
+  isDescendantOf(element: File): boolean {
+    let parent = this.parent;
+    if (!parent) {
+      return false;
+    }
+    while (parent.parent) {
```
mbebenita: I think this fails for `file.isDescendantOf(file.parent)`, if `file.parent.parent` is `null`.

ollelauribostrom: Yes you are right. Would this be a better approach?

```javascript
  isDescendantOf(element: File): boolean {
    let parent = this.parent;
    while (parent) {
      if (parent === element) {
        return true;
      }
      parent = parent.parent;
    }
    return false;
  }
```

Also, any pointers on where to create the test case? Can I create a new file called file.spec.tsx in tests/utils?

mbebenita: You can probably just add a file to the `tests/unit` directory. Create a basic directory structure and check that `isDescendentOf` works correctly.

tests/unit/descendant.spec.ts
```diff
@@ -0,0 +1,30 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+import { Directory } from "../../src/model";
+
+function getDirectoryStructure() {
+  const a = new Directory("test");
+  const b = a.newDirectory("/b");
```
mbebenita: No need to add the `/` to the name.

yurydelendik:
Thank you for the patch!

PR #139, opened 2018-02-25
Add parameter to open a template, just like a fiddle

dcodeIO:
### Summary of Changes

* Added a parameter `template=` / `t=` (like `fiddle=`/ `f=`) that opens a template, either by its full name or template directory.
* Now also opens the "New Project Dialog" if a fiddle or template does not exist.

### Test Plan

* Open WebAssembly Studio with `?t=hello_world_c`, `?template=Hello World in C` etc.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -89,11 +89,13 @@ import { Errors } from "../errors";
 import { ControlCenter } from "./ControlCenter";
 import Group from "../utils/group";
 import { StatusBar } from "./StatusBar";
+import fetchTemplates from "../utils/fetchTemplates";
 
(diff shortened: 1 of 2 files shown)


mbebenita:
I think we should unify the concepts of fiddles and templates. If we made templates behave just like fiddles then we wouldn't need two different code paths. What if we create a npm task that generates fiddle JSON objects from template directories? We could also get rid of the `templates.js` file. @tschneidereit what do you think?

AnshulMalik:
That would be great @mbebenita

mbebenita:
The infrastructure should be in place to make this easy now.

PR #150, opened 2018-03-03
Fix various build issues (on Windows)

dcodeIO:
### Summary of Changes

* Fix bin files being non-executable through CRLF on Windows
* Fix invokation of bin files in package.json to be sure
* Upgrades webpack to v4 to(/and?) get rid of UglifyJS breaking the build
* Upgrades other dependencies as well to be sure

### Test Plan

* Run all the npm scripts on Windows
* Launch WebAssembly Studio and check that nothing broke

As a side note, when I navigate to https://webassembly.studio, there are no templates atm. Not sure if that's somehow connected. Works locally for me now, though.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+bin/bundle-templates.js text eol=lf
+bin/check-bundle-size.js text eol=lf
--- a/bin/bundle-templates.js
+++ b/bin/bundle-templates.js
@@ -20,7 +20,7 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
- 
+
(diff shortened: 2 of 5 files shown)


mbebenita:
> no templates atm

The reason for this is because for some reason, `bundle-templates` is not executed.

dcodeIO:
> The reason for this is because for some reason, bundle-templates is not executed.

I didn't double check my theory, but it might be because it is `&& ./bin/build-templates.js` in the `build` script, with the previous one failing on uglify but still emitting a bundle, so the `&&` step isn't executed.

mbebenita:
Let's try it out, merging this.

dcodeIO:
Doesn't seem to fix it. Might it be that the build-bot is just running `webpack` instead of `npm run build`? Just guessing.

Issue #158, opened 2018-03-03
Inconsistent 'modified' indicator


chrisjj:
'Modified' is indicated on the file tab by a circle, yet in the Project Workspace by an 'M'.

![image](https://user-images.githubusercontent.com/6157073/36938957-713602d6-1f21-11e8-827e-dfbed77eb035.png)

garvitdelhi:
Hi, I would like to pick this up, I am done with setting up the dev env and also have started to work on it.

garvitdelhi:
@mbebenita can you review this?

PR #160, opened 2018-03-04
Made modifier indicator consistent with tab representation

PhilDenoncourt:
Associated Issue: #158 

### Summary of Changes

* Updated styling so that modified indicator was the same in the directory tree as in the tab.

### Test Plan

Tell us a little a bit about how you tested your patch.

Example test plan:

-Create a new project
-Modify a file in src
-Notice the circular indicator
modifyindicator

### DIFF
--- a/style/global.css
+++ b/style/global.css
@@ -861,17 +861,14 @@ a {
 }
 
 .monaco-icon-label.dirty::after {
-  content: "M";
-  font-size: 10px;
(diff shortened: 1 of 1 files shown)


mbebenita:
This used to be the way we marked files in the workspace view as being modified. When we added the transient flag (marked with a `T`) we experimented with marking modified files with `M`. I don't mind leaving the dot, but we should see how it interacts with other markers.

PhilDenoncourt:
Perhaps I misunderstood the issue?  I thought the desire was to change the M to the dot.  The transient indicator continues to show as T.

mbebenita:
The other option would be to change the dot to an `M` in the tab view.

RMCoder198:
What will your code show if file is transient and modified ?

You have deleted the code that shows M T when file is transient and modified.

Issue #66, opened 2018-02-12
Deleting a file that is opened in a tab doesn't close the tab.


mbebenita:
Tabs should not hang on to deleted files.

hleal18:
Hello, I'm getting familiar to the project and I'd like to know if there's someone working on this bug already, otherwise I'd like to give it a try and work on it.

mbebenita:
I don't think anyone is working this at the moment, feel free to give it a shot.

yurydelendik:
Fixed by #161

PR #161, opened 2018-03-05
Fix opened tabs of deleted files

hleal18:
Associated Issue: #66 

### Summary of Changes

* Created an Action that specifically evaluates every view of every group looking for the ones that contains the file that was deleted by the user. The action *closeTabs* is invoked along *deleteFile*.
* A private method on AppStore was created. Is a refactored version of a piece of code that *closeView* originally had. Now *closeView* and *closeTabs* invoke *closeGroup* when a group needs to be closed and also handles the next appropriate tab to focus.

### Test Plan

It was tested by opening the same file two or three times by splitting the editor. Then, the file was deleted and all the tabs related to that file were closed.


yurydelendik:
Thank you for the patch.

PR #164, opened 2018-03-09
WIP: Node Editor

mbebenita:
This PR starts adding support for editing node graphs. Node graphs can be used to write up modules in interesting ways. The backing store for node graphs are JSON files and much of this code follows the same architecture as the Markdown preview.

This is sill very much WIP.

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -27,6 +27,7 @@ SOFTWARE.
         <title>WebAssembly Studio</title>
         <link href="style/global.css" rel="stylesheet">
         <link href="style/theme.css" rel="stylesheet">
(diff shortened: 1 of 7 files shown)


mbebenita:
* Added port snapping.
* Removed editable ports (will re-add them as external widgets that sit on the side of the port and outside of the node itself.
* Made input and output ports appear on the same horizontal line.
* Added an edge slicing tool.
* Added panning and zooming, although node dragging is currently broken.
* Made edges straight, this makes it cleaner and easier to do edge / line intersections.
. 
image

Issue #162, opened 2018-03-06
When forking, or sharing, suggest code formatting.


mbebenita:
Suggest to the user that they should format their code before sharing it with others.

PR #165, opened 2018-03-09
Message to, fork a formatted code

RMCoder198:
Associated Issue: #162 

 Summary of Changes

Whenever user forks a project , user will receive a message to fork a formatted code

Test Plan

Make changes in file

Run localhost 

check the desired changes

run command npm run test

Screenshots/Videos (OPTIONAL)
fork

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -295,6 +295,7 @@ export class App extends React.Component<AppProps, AppState> {
   async fork() {
     const projectModel = this.state.project.getModel();
     pushStatus("Forking Project");
+    this.toastContainer.showToast(<span>"Fork a formatted code for better readability"</span>);
     const fiddle = await Service.saveProject(projectModel, []);
     popStatus();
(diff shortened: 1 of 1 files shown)


mbebenita:
This is a bit tricky from an UX perspective. Really what you want is that whenever someone wants to share something, you want them to be reminded they should format their code. Maybe the easiest thing is to put a warning on the sharing page reminding the user to format their code.

RMCoder198:
@mbebenita 
Only forked code gets  shared , that's why i chose to remind user to format  code when user  forks his/her code.
But if you want to do it in your way , that is good .

mbebenita:
> remind user to format code when user forks his/her code

The issue here is that I think you want to remind them, and prevent them from actually forking if they decide to go back an format it. If you let them fork, and then give them a reminder, then they'll have to go back and fork again after they formatted things.

RMCoder198:
Ok ,i got it.

PR #168, opened 2018-03-13
Fixed pinning issue

myerich:
Associated Issue: #57

### Summary of Changes

When a user double clicks a tab other than the current preview tab, the preview tab is no longer incorrectly pinned.

### Test Plan

Made changes to the group.ts file
Confirmed results on a local dev server
ran `npm run test` and passed all tests


yurydelendik:
Thank you for the patch

PR #172, opened 2018-03-21
Fix silent overwrite behaviour when moving file/folder

ollelauribostrom:
Currently, when moving a file/folder into a folder already containing a file/folder (immediate child) by that name - the moved content is silently deleted which may lead to the user loosing work. This PR fixes this behavior.

### Summary of Changes
* Adding additional check in onDragOver (DirectoryTree.tsx) making the accept property return false if targetElement already contains immediate child with same name as file/folder being moved

### Test Plan
- Create empty C project
- Create new folder /test and /test/src
- Create new file at root called main.c
- Try moving /main.c -> /src (Should no longer be accepted)
- Try moving /src -> /test (Should no longer be accepted)


mbebenita:
Thanks for the patch. I've thought about this problem before. A similar problem exists when uploading files with the same name. I think most UIs will rename the dragged / uploaded file to `main.c.2` instead of preventing the drag. But for now, I think this is a good solution.

PR #173, opened 2018-03-21
Added StatusBar test

ollelauribostrom:
### Summary of Changes
* Added tests for StatusBar component

### Test Plan
- Checks that StatusBar renders an empty string if project does not have a status
- Checks that StatusBar does not apply the active className if project does not have a status
- Checks that StatusBar update text and className on status change

### DIFF
--- a/tests/components/StatusBar/StatusBar.spec.tsx
+++ b/tests/components/StatusBar/StatusBar.spec.tsx
@@ -1,0 +1,33 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+import * as React from "react";
+import {shallow} from "enzyme";
+import {StatusBar} from "../../../src/components/StatusBar";
+import {initStore} from '../../../src/actions/AppActions';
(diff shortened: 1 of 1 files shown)


tests/components/StatusBar/StatusBar.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import * as React from "react";
+import {mount} from "enzyme";
+import {StatusBar} from "../../../src/components/StatusBar";
+import {initStore} from '../../../src/actions/AppActions';
+import appStore from '../../../src/stores/AppStore';
+
+initStore();
+const project = appStore.getProject().getModel();
+const wrapper = mount(<StatusBar/>);
```
yurydelendik: can you move this out of module global to function inside "tests for statusbar"?  it is expected that the mount will be called only when an "it" body is called.

ollelauribostrom:
Updated this. Now using shallow rendering since the React lifecycle methods apparently are being called (As of Enzyme v3).

yurydelendik:
Thank you for the patch.

Issue #174, opened 2018-03-22
wat project: build error causes app stuck at "building"


kenrick95:
1. https://webassembly.studio/?f=qwejfuumnhd
2. Notice that there is error in wat codes (`$add` is not defined); Note: this is intentional
3. Click "Build"
4. App will stuck at "Building project..." and no other action could be made

<img width="724" alt="screen shot 2018-03-22 at 12 03 01 pm" src="https://user-images.githubusercontent.com/3090380/37750636-0ca55306-2dc9-11e8-89e7-8712cbdb0c23.png">

The error at browser console is:
```
Error: resolveNames failed:
test.wat:6:23: error: undefined function variable "$add"
  (export "add" (func $add))
                      ^^^^
```

---

Update: The issue here is that if the build process failed (in this case due to undefined function variable), the WebAssemblyStudio app is stuck at "building"

mbebenita:
This is really strange. If i create an empty .wat project, the main.wat file looks different. The name of the main function is $add, not $error.

kenrick95:
Yeah, that's my point, if I mistakenly have an error in my .wat project, in this case naming the function as `$error` instead of `$add`, seems like the whole WebAssemblyStudio app will crash (and I need to reload the page to recover its functionality, losing the codes that I've worked on)

yurydelendik:
Fixed by #176

PR #175, opened 2018-03-22
Listen to worker error event

kenrick95:
This is to throw the error back at main thread instead of silently failing at worker script

Fixes: #174

### Summary of Changes

* At service.ts, right now, when it spawns a worker, it does not handle if there are error thrown at worker script. Hence, #174 could happen.
* The change in this PR, is to make service.ts listen to the worker error event, and reject the postMessage promise if there are error happening at worker event.

### Test Plan

- Following the steps listed in #174 no longer makes the editor stuck at "building"

### Screenshots

Instead, it shows the error captured as output in "Output" window

capture-error-captured

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -138,6 +138,8 @@ async function getServiceURL(to: ServiceTypes): Promise<string> {
 class ServiceWorker {
   worker: Worker;
   workerCallbacks: Array<{fn: Function, ex: Function}> = [];
+
+  workerErrorCallback: Function;
   nextId = 0;
   private getNextId() {
     return String(this.nextId++);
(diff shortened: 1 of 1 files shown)


mbebenita:
Thanks for this patch. I was trying to solve the problem by having the worker send a message on workers. I'll land that technique as part of #176 but I'll also rebase this and land it to handle the general case when an unhandled exception is thrown in the worker thread.

src/service.ts
```diff
@@ -152,10 +154,17 @@ class ServiceWorker {
       if (e.data.success) {
         cb.fn(e);
       } else {
+        this.workerErrorCallback = null;
         cb.ex(e);
       }
       this.workerCallbacks[e.data.id] = null;
     });
+    this.worker.addEventListener("error", (e) => {
```
yurydelendik: This shall never happen, unless there some critical error with worker. From the descriptions of issue and PR, I cannot understand what causes it and why we cannot fix it on worker's processMessage side.

kenrick95:
Since #176 is merged, I guess this PR is no longer relevant, as the error handling has been done as part of that PR.

Issue #99, opened 2018-02-16
Just-opened files are all-selected occasionally


dcodeIO:
![select](https://user-images.githubusercontent.com/1136893/36300227-ad9a250e-1300-11e8-9b8c-aadc8d3dfbad.jpg)

Unfortunately, I haven't yet figured out a reliable way to trigger this, but this happens rather regulary on my end when switching between files. Did someone else notice this or is it my PC? :)

haxxorsid:
Not me.

yurydelendik:
@dcodeIO what do you use for switching between files? with exact steps, including e.g. "click shift key + holding middle mouse button".

dcodeIO:
I think I found out how to reproduce it:

* Create a new project
* Close all files
* Left-click the empty editor area once (where the file contents would be otherwise)
* Open a file using the directory tree

haxxorsid:
Step 3: Double left clicking the empty editor area reproduces it.

mbebenita:
Is this happening in both Firefox and Chrome?

dcodeIO:
Seems that in Firefox, a single left click is sufficient to trigger this, while in Chrome, a double left click is required.

hleal18:
I want to know if there is someone working on this issue, otherwise, I'd like to work on it.

haxxorsid:
@hleal18 nope I think. You're welcomed to work on this.

yurydelendik:
Fixed by #179

PR #179, opened 2018-04-04
Fix all-content-selection when opening or switching tabs

hleal18:
Associated Issue: #99 

### Summary of Changes

* Remove the browser content selection when a different tab than the current one is clicked.
* Remove the browser content selection when a tab is opened by clicking a file.

This two changes avoids the propagation of the browser content selection when navigating tabs and it doesn't affect the text selection that the Monaco editor does on the code.

## Chrome:
Although the plan to reproduce it on chrome is explained in the issue, I couldn't do it exactly in that way. However, the strange behavior of all-content-selection can be reproduced by selecting the content of a markdown document on preview mode by clicking three times on it, and then, changing to the other tabs to observe how all the content-selection is propagated to them.

## Mozilla:
This was reproduced similar as it is explained on the issue, although it required me to click two times on an empty space, and then open a file to observe how is all selected.

With the patch, neither of this two cases occur.

Example test plan:

- [x] Create a new project with a default template.
- [x] Check that the bug is solved by trying to reproduce it with the steps mentioned above and in the issue referenced.

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -495,6 +495,10 @@ export class App extends React.Component<AppProps, AppState> {
           }}
           onChangeViewType={(view, type) => setViewType(view, type)}
           onClickView={(view: View) => {
+            if (!(appStore.getActiveTabGroup().currentView === view)) {
+              // Avoids the propagation of content selection between tabs.
+              window.getSelection().removeAllRanges();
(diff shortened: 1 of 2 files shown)


yurydelendik:
Thank you for the patch

PR #180, opened 2018-04-06
Added NewDirectoryDialog test

ollelauribostrom:
### Summary of Changes
* Added tests for NewDirectoryDialog component
* Changing title of cancel button

### Test Plan
- Checks that it renders the correct elements
- Checks that it displays error & disables create button on already existing name
- Checks that it displays error & disables create button on invalid name
- Checks that it invokes onCancel callback when clicking cancel button
- Checks that it invokes onCreate callback (with created directory) when clicking create button


yurydelendik:
Thank you for the patch

Issue #91, opened 2018-02-15
Remove/reduce forceUpdate usages


yurydelendik:
There are several components that use `this.forceUpdate()`:

- [x] App.tsx
- [ ] ~~EditorPane.tsx~~
- [x] StatusBar.tsx
- [x] Binary.tsx
- [x] ControlCenter.tsx
- [x] Markdown.tsx
- [x] UploadFileDialog.tsx
- [x] Viz.tsx
- [x] ViewTabs.tsx
- [x] Workspace.tsx

It is [not-recommended](https://reactjs.org/docs/react-component.html#forceupdate) to use this method to trigger `render`. We need to:

a) document/comment a reason this method is called,
b) or, replace it with `shouldComponentUpdate`.

PR #203, opened 2018-04-20
Remove forceUpdate from src/components/UploadFileDialog.tsx

yurydelendik:
Associated Issue: #91

### Summary of Changes

* Add timestamp attribute to File model
* Use onDidChildrenChange on timestamp update
* Remove forceUpdate from UploadFileDialog.tsx


yurydelendik:
More design work needed around updating buffer/name/etc.

PR #204, opened 2018-04-23
C-Template: Create function prototype for main()

LeonardKoenig:
Associated Issue: N//A

### Summary of Changes

* An empty parameter-list in a function declarator does not specify the parameters, except when part of a function definition, however in either case it does not create a function prototype, which allows for calling `main()` with more than zero arguments. This is a primarily stylistic issue here, but it's good to not have an example show "bad" code.

### Test Plan

N/A

N/A

I didn't bother with creating a feature branch due to the small impact of the change.


yurydelendik:
thank you

PR #209, opened 2018-05-01
Support template file with external contents

tschneidereit:
Associated Issue: None

### Summary of Changes

* Add support for loading a template's files on-demand instead of embedding them all into the templates file itself.
* The code that creates the templates file is also changed to copy files over to `dist/templates/[template name]`.
* Same for the code for loading files to properly resolve paths relative to the templates file.

### Test Plan

- [x] Open WebAssembly.studio
- [x] Select a template
- [x] Click the "Create" button
- [x] Open all of the project's files to ensure they're all as expected
- [x] Build and Run project to ensure it works as expected
- [x] Repeat for all templates

### DIFF
--- a/bin/build-templates.sh
+++ b/bin/build-templates.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+rm -r dist/templates &> /dev/null
+cp -r templates dist/
+node ./bin/bundle-templates.js templates dist/templates
--- a/bin/bundle-templates.js
+++ b/bin/bundle-templates.js
@@ -24,6 +24,9 @@
 const path = require("path");
 const fs = require("fs");
 
+const templatesDir = process.argv[2];
+const outputPath = process.argv[3];
+
 function walk(base, callback) {
   let files = fs.readdirSync(base);
   files.forEach(file => {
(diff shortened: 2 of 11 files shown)


tschneidereit:
@yurydelendik, @mbebenita this is the change I was talking about earlier. It should make it much easier to provide custom templates via an embedding. We probably need to talk about security implications of that before actually making that possible.

tschneidereit:
I'll look into the test failure tomorrow.

tschneidereit:
@ollelauribostrom, @yurydelendik tells me you know a lot about the test infrastructure, so perhaps you can help me with this test failure? `fetch` doesn't just work in tests, so I added a mock for `getConfig`, which is now used by the code under test. However, now I'm running into issues with `URL`. Can you give me a pointer for how to work around those? I don't even need the result of what `URL` is used for here, but refactoring the code to only use `URL` if it's really needed would make it much less clear :(

ollelauribostrom:
@tschneidereit I'll look in to this tonight!

tschneidereit:
@ollelauribostrom, excellent, thank you so much!

yurydelendik:
Thank you for the patch!

PR #216, opened 2018-05-15
Implement Service.import to be used in the build.ts files.

yurydelendik:
Associated Issue: ability to use ES6 module import function in the build script

### Summary of Changes

* Extract and move runTask and contextify
* Move task evaluation into iframe global, and fix empty_ts project
* Implement Service.import 
* Provide sample code in the  misc/arc-templates/arc_wat/build.ts


misc/arc-templates/arc_wat/build.ts
```diff
@@ -8,6 +8,13 @@ gulp.task("build", async () => {
 });
 
 gulp.task("publish", async () => {
+    // FIXME send only data (sample of Service.import call)
```
tschneidereit: We actually want to send both the data and the files, but I'll work on that.

misc/arc-templates/arc_wat/build.ts
```diff
@@ -8,6 +8,13 @@ gulp.task("build", async () => {
...
+    const rows = 30, cols = 40, frameCount = 50, fps = 10;
+    const { transform } = await (await Service.import('src/module.js')).default();
+    const buffer = new ArrayBuffer(cols * rows * frameCount * 3);
+    transform(buffer, rows, cols, frameCount, fps, true);
+    logLn('data > ', new Uint8Array(buffer));
```
tschneidereit: Nit: remove logging

Issue #220, opened 2018-05-16
Multiple Monaco editor dependencies


ollelauribostrom:
There are currently two Monaco Editor dependencies. One is located in lib/monaco-editor and one is listed in package.json. These two are identical when it comes to the versions, so we could try to remove one of them.

In my opinion it would probably be easiest to manage the Monaco Editor dependency if sticking to the one installed from npm. To do this, Webpack must be configured to load the editor as described in ["Integrating the ESM version of the Monaco Editor"](https://github.com/Microsoft/monaco-editor/blob/HEAD/docs/integrate-esm.md).

Any thoughts?

yurydelendik:
@mbebenita mentioned that he did some cosmetic modifications to the files

ollelauribostrom:
I've experimented a bit with this in #221. Looks like it could be doable since the Monaco Editor distributed at npm now includes the components TreeView, ContextMenuService, ContextViewService, TreeDefaults and Action that are used in monaco-utils.ts.

mbebenita:
Great, then it's definitely the right way to go.

yurydelendik:
Closed by #221

PR #221, opened 2018-05-17
Removing lib/monaco-editor dependency

ollelauribostrom:
Associated Issue: #220

### Summary of Changes
* Removed the lib/monaco-editor dependency
* Updated the webpack config to load the monaco editor from node_modules instead

### Todo
- [x] ~~Fix issue with the context menu~~ (Fixed by supplying a logService when constructing MonacoUtils.ContextViewService in DirectoryTree.tsx and Problems.tsx)
- [x] ~~Fix issue in taskRunner.ts (Occurs on build/run)~~ (Fixed by adding require.js)
- [x]  ~~Fix failing tests (Modify the monaco-editor mock)~~
- [x] ~~Investigate line 45 in src/utils/registerLanguages.ts (Commented out to make this run)~~ (Fixed by adding `Wat.HoverProvider as any`)
- [x] ~~Investigate line 198 in src/stores/AppStore.ts (Removed identifier to make this run)~~ ([API change](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.iidentifiedsingleeditoperation.html))
- [x] ~~Investigate bundle sizes~~


yurydelendik:
The loading via `await import` looks like working, just producing lots of bundles and takes longer to compile. Can we use top-module-imports for monaco inside single file that will be async-imported be src/monaco-utils.ts to avoid bundle pollution?

I'm also wondering why worker is trying to import react stuff. Do we have some cyclic dependency so we pull a world into the worker?

ollelauribostrom:
Also noticed this, will investigate further @yurydelendik

ollelauribostrom:
Modified the webpack config a bit, resulting in a single monaco-editor.bundle.js (1.34mb Gzipped size) according to the Webpack Bundle Analyzer.

ollelauribostrom:
@yurydelendik 

This now looks like this:

skarmavbild 2018-05-24 kl 11 19 37

And the process is basically:
1. Import the editor chunk
2. Register the theme
3. Render the App component
4. Import the languages chunk
5. Register the languages

https://github.com/ollelauribostrom/WebAssemblyStudio/blob/feb0093f8ad9b390db27e78a8e539faba2c1559f/src/index.tsx#L107

yurydelendik:
Nice! Thank you for the patch.

PR #222, opened 2018-05-18
hello_world_rust: fix README.md

srenatus:
It said main.rs, where it should say lib.rs to be accurate.

Trivial, so please bear with me in foregoing the PR template 😃

### DIFF
--- a/templates/hello_world_rust/README.md
+++ b/templates/hello_world_rust/README.md
@@ -9,7 +9,7 @@ bundler so you can poke around all the raw output!
 
 Some files you may be interested in are:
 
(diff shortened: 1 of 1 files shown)


srenatus:
> Another idea (without at all being familiar with any naming conventions in the rust community) could be to just rename lib.rs to main.rs

I'm only vaguely familiar with those conventions myself, but I think `main.rs` is commonly used for binaries, whereas `lib.rs` is the filename of choice for libraries. So, in this case, I think `lib.rs` is preferable.

yurydelendik:
Thank you for the patch.

PR #224, opened 2018-05-18
Fixes #190 - Clipped Monaco context menu

### LINKED ISSUE #190: context menu in source pane is often obscured by output pane
I tried the inline studio window in https://hacks.mozilla.org/2018/04/sneak-peek-at-webassembly-studio/ . It works well, but if I right-click on a source term towards the bottom of the source pane, the context menu that appears is obscured by the output pane below. In the screenshot, you only see the first menu item "Change all occurrences", the rest (Cut, Copy, Command Palette, etc.) are obscured. It looks like the z-index of the div `context-view monaco-menu-container bottom left` is only 8, I tried adjusting it to make it appear on top without success.
screenshot-2018-4-12 sneak peek at webassembly studio mozilla hacks - the web developer blog

Context menus can still be clipped by the bottom of the Studio window. Perhaps the UI toolkit you're using doesn't ensure all pop-up menu items are accessible by repositioning them or adding a scrollbar.

This is in Firefox Nightly 61.0a1 Build ID 20180410220129 on Windows 10.

ollelauribostrom:
Associated Issue: #190 

### Summary of Changes
* Uses fixed positioning for the context menu css
* Registers an event listener (onContextMenu) when the editor is created
* Each time the context menu is opened, the correct position is calculated and set

### Test Plan
* Create C "hello, world" project
* Open a file
* Right click to open the context menu
* The context menu should no longer get clipped

### Screenshots
skarmavbild 2018-05-18 kl 14 16 24

### DIFF
--- a/src/components/editor/Editor.tsx
+++ b/src/components/editor/Editor.tsx
@@ -149,6 +149,12 @@ export class Monaco extends React.Component<MonacoProps, {}> {
       this.container.removeChild(this.container.lastChild);
     }
     this.editor = monaco.editor.create(this.container, options as any);
+    this.editor.onContextMenu((e: any) => {
+      const anchorOffset = { x: -10, y: -3 };
+      const menu: HTMLElement = document.querySelector(".monaco-editor > .monaco-menu-container");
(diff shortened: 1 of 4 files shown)


yurydelendik:
Thank you for the patch.

PR #227, opened 2018-05-18
Add uploading of files via drag and drop.

mbebenita:
This makes it possible to drag files from the local file system into the project. At the moment, this is limited to the files for which we have mime types. Due to differences between Firefox and Chrome, this PR has some hacky workarounds to make both browsers work.

drag


yurydelendik:
Thank you for the patch.

PR #230, opened 2018-05-22
Fixing context menu issues

ollelauribostrom:
Related to #224. Fixes some additional issues with the context menu.

### Summary of Changes
* Disables editor scroll when the context menu appears to make the menu itself scrollable
* Calculates the menus max height to avoid overflowing outside the window
* In the onContextMenu callback the menu element is now selected on this.container to ensure that we change styles on the correct element

### Test Plan
- Create empty C project
- Open a file
- Right click to open context menu
- The context menu should now be scrollable
- The context menu should no longer be able to overflow outside of the window


ollelauribostrom:
Build needs to be restarted, seems like it failed due to an [npm incident this morning](https://status.npmjs.org/incidents/b307lbjrdql7)

yurydelendik:
Thank you for the patch

PR #235, opened 2018-05-28
Change Arc templates to swap columns and rows

tschneidereit:
The website currently misinterprets columns and rows, and we don't want people to have to do the same in their modules, so we swap them before sending the animation off to the website.


tschneidereit:
I ran tests locally, and this only affects the Arch website embedding, so I'll go ahead and land this.

PR #236, opened 2018-05-28
Adding jest-enzyme as dev-dependency

ollelauribostrom:
Adding [jest-enzyme](https://github.com/FormidableLabs/enzyme-matchers) as dev-dependency. This adds a bunch of assertions that will make the tests easier to understand / yield better error messages.

Add `import "jest-enzyme";` in the test file to import typings


yurydelendik:
Thank you

PR #239, opened 2018-05-29
Improve toolbar buttons and shortcuts for Arch embedding

tschneidereit:
Adds a "Build & Preview" button for the Arch website embedding, and changes the `CmdCtrl + Enter` and `CmdCtrl + Alt + Enter` shortcuts to work properly for that embedding.


tschneidereit:
Change only affects Arch embedding, merging without review.

PR #244, opened 2018-05-30
Added AppActions tests

ollelauribostrom:
### Summary of Changes
* Added tests for AppActions.ts


yurydelendik:
Thank you for the patch.

Issue #184, opened 2018-04-11
Use kiwi.js instead of Cassowary?


JobLeonard:
Cassowary.js hasn't been updated in years.

When someone commented on that in the github repo, [the maintainers themselves suggested kiwi.js](https://github.com/slightlyoff/cassowary.js/issues/79) for an alternative faster implementation:

https://www.npmjs.com/package/kiwi.js

I don't think Cassowary is currently a bottleneck, but if the API is identical and it doesn't introduce new bugs, then why not get the free performance boost?

mbebenita:
@JobLeonard thanks for the tip, I hadn't come across this implementation.

ollelauribostrom:
Gave kiwi.js a try this weekend. For the most part, the API is the same but there are some differences which made me fail in swapping out this dependency. 

I could not find any kiwi.js equivalence for: 
* solver.addStay(...)
* Cassowary.Equation(..)
* Cassowary.Inequality(..)

However, i'm not too familiar with the solving algorithm. Does @mbebenita know if there are any way around those differences?

mreinstein:
> Does @mbebenita know if there are any way around those differences?

You could submit improvements to kiwi.js directly :) My impression is the maintainer is approachable and reasonable from my limited interactions.

mreinstein:
All 3 of the aforementioned bullets have equivalents in kiwi.js see https://github.com/IjzerenHein/kiwi.js/issues/17 for details on these.

Also, shameless plug, if you're looking for something a little higher level where you can specify a set of equations as a string, I wrote/maintain `constraint-solver` npm module, which layers on top of the latest kiwi.js: https://github.com/mreinstein/constraint-solver

PR #248, opened 2018-05-31
Adding tests for Split component

ollelauribostrom:
### Summary of Changes
* Added tests for the Split component
* Removing unused utility functions
* Removing unused static EventDispatcher
* Moving out utility functions to own file

After #184  is closed, the actual solving can be tested in further detail.

### DIFF
--- a/src/utils/splitUtils.ts
+++ b/src/utils/splitUtils.ts
@@ -1,0 +1,35 @@
+/* Copyright 2018 Mozilla Foundation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


src/utils/splitUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+  return (x | 0) + "px";
+}
+
+export function assignObject(to: any, from: any) {
+  for (const x in from) {
+    if (!(x in to)) {
+      to[x] = from[x];
+    }
+  }
+  return to;
```
yurydelendik: The behavior look identical to `return {...from, ...to};` -- we probably want to remove assignObject

yurydelendik:
lgtm. thank you for the patch.

PR #252, opened 2018-05-31
remove npm install --dev instruction from readme

ashleygwilliams:


tschneidereit:
@ashleygwilliams just explained this to me: the `--dev` flag has been deprecated for a long time and doesn't do anything useful, so we can just remove it.

PR #253, opened 2018-05-31
Provide full path for files in Service.compileFiles

yurydelendik:
Provides full path for files that are send to the service. Test is provided.

### DIFF
--- a/tests/utils/service.spec.tsx
+++ b/tests/utils/service.spec.tsx
@@ -1,0 +1,55 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+import { Project, Directory, File, FileType } from "../../src/model";
+import { Service, Language } from "../../src/service";
+import * as compilerServices from "../../src/compilerServices";
+
+class TestService {
+  public input;
+  async compile(input) {
+    this.input = input;
+    const items = { "a.txt": { content: "text" } };
+    return { success: true, items, };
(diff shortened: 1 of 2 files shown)


tests/utils/service.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+    return { a, b, c, };
+}
+
+describe("Service.compileFiles tests", () => {
+  let saveCCS;
+  beforeAll(() => {
+    saveCCS = (compilerServices as any).createCompilerService;
+  });
+  afterAll(() => {
+    (compilerServices as any).createCompilerService = saveCCS;
```
yurydelendik: @ollelauribostrom I am not sure if there is something better than doing restoring mocks this way.

ollelauribostrom: You could probably use jest.mock instead, so something like,

```js
const compile = jest.fn();

jest.mock("../../src/compilerServices", () => ({
  createCompilerService: () => ({
    compile: compile.mockImplementation(() => ({
      success: true,
      items: { "a.txt": { content: "text" } }
    }))
  }),
  Language: {
    JavaScript: "javascript"
  }
}));

import { Project, Directory, File, FileType } from "../../src/model";
import { Service, Language } from "../../src/service";
```

```js
expect(result).toEqual(...)
expect(compile).toHaveBeenCalledWith(...)
```

```js
afterAll(() => {
  jest.unmock("../../src/compilerServices");
});
```

PR #254, opened 2018-05-31
Cargo compilation support

skade:
Work in progress PR to make the RustService compile using the /cargo endpoint

### DIFF
--- a/misc/arc-templates/arc_rust/build.ts
+++ b/misc/arc-templates/arc_rust/build.ts
@@ -2,10 +2,13 @@ import * as gulp from "gulp";
 import { Service, Arc, project, logLn } from "@wasm/studio-utils";
 
 gulp.task("build", async () => {
-    const options = { lto: true, opt_level: 's', debug: true };
-    const data = await Service.compileFile(project.getFile("src/lib.rs"), "rust", "wasm", options);
+    const options = { debug: true, cargo: true };
+    const data = await Service.compileFiles([
+        project.getFile("src/lib.rs"),
+        project.getFile("Cargo.toml")            
+    ], "rust", "wasm", options);
     const outWasm = project.newFile("out/lib.wasm", "wasm", true);
-    outWasm.setData(data);
+    outWasm.setData(data["a.wasm"]);
 });
 
 gulp.task("publish", async () => {
--- a/src/@types/tar-js/index.d.ts
+++ b/src/@types/tar-js/index.d.ts
@@ -1,0 +1,6 @@
+// Type definitions for [tar-js]
+// Project: [WebAssemblyStudios]
+// Definitions by: [Florian Gilcher] <flo@andersground.net>
+
+declare module '*';
+
--- a/src/compilerServices/rustService.ts
+++ b/src/compilerServices/rustService.ts
@@ -22,17 +22,40 @@ import { CompilerService, ServiceInput, ServiceOutput, Language } from "./types"
 
 import { sendRequestJSON, ServiceTypes } from "./sendRequest";
 import { decodeBinary } from "./utils";
+import * as Tar from "tar-js";
+import { base64EncodeBytes } from "../util";
 
 export class RustService implements CompilerService {
   async compile(input: ServiceInput): Promise<ServiceOutput> {
-    const files = Object.values(input.files);
-    if (files.length !== 1) {
-      throw new Error(`Supporting compilation of a single file, but ${files.length} file(s) found`);
-    }
-    const [ fileRef ] = Object.keys(input.files);
-    const code = files[0].content;
+    let result;
+    let fileRef;
     const options = input.options;
-    const result = await sendRequestJSON({ code, options, }, ServiceTypes.Rustc);
+    const cargo = options["cargo"];
+
+    if (cargo) {
+      const tarBuffer = new Tar();
+
+      const files = input.files;
+      Object.entries(files).forEach(
+        ([name, file]) => {
+          tarBuffer.append(name, file.content, {});
+        }
+      );
+
+      const tar = base64EncodeBytes( tarBuffer.out );
+
+      result = await sendRequestJSON({ tar, options, }, ServiceTypes.Cargo);
+      fileRef = Object.keys(files)[0];
+    } else {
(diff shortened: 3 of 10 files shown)


misc/arc-templates/arc_rust/build.ts
```diff
@@ -1,9 +1,20 @@
...
+        }
+        if (file.name.endsWith(".rs")) {
+            files.push(file);
+        }
+    }, false, true);
+
+    const options = { lto: true, opt_level: 's', debug: true, cargo: true };
+    const data = await Service.compileFiles(files, "rust", "wasm", options);
     const outWasm = project.newFile("out/lib.wasm", "wasm", true);
     outWasm.setData(data);
```
yurydelendik: outWasm.setData(data["a.wasm"]);

src/service.ts
```diff
@@ -240,6 +240,76 @@ export class Service {
...
+    const service = await createCompilerService(from, to);
+
+    const fileNameMap: {[name: string]: File} = files.reduce((acc: any, f: File) => {
+      acc[f.name] = f.getPath();
+      return acc;
+    }, {} as any);
+
+    const input = {
+      files: files.reduce((acc: any, f: File) => {
+        acc[f.getPath()] = {
```
yurydelendik: Use `getProjectFilePath()` here https://github.com/wasdk/WebAssemblyStudio/blob/master/src/service.ts#L87

src/service.ts
```diff
@@ -289,6 +359,7 @@ export class Service {
     return outputFiles;
   }
 
+
```
yurydelendik: nit: unneeded change

src/compilerServices/rustService.ts
```diff
@@ -22,17 +22,40 @@ import { CompilerService, ServiceInput, ServiceOutput, Language } from "./types"
...
+      let files = input.files;
+      Object.entries(files).forEach(
+        ([name, file]) => {
+          tarBuffer.append(name, file.content, {})
+        }
+      );
+
+      let tar = base64EncodeBytes( tarBuffer.out );
+
+      result = await sendRequestJSON({ tar, options, }, ServiceTypes.Rustc);
```
yurydelendik: We need to add ServiceTypes.Cargo here (and use different config)

skade:
@yurydelendik the last patch should fix your review points. Thanks! :)

src/service.ts
```diff
@@ -240,6 +240,76 @@ export class Service {
...
+    }
+
+    const resultItemsIterable = {
+      [Symbol.iterator]: function*() {
+        if (!result.items) {
+          return;
+        }
+        for (const name of Object.keys(result.items)) {
+          yield {name, item: result.items[name]};
+        }
```
tschneidereit: This can be
```js
for (const [name, item] of Object.entries(result.items)) {
  yield {name, item};
}
```

src/compilerServices/rustService.ts
```diff
@@ -22,17 +22,40 @@ import { CompilerService, ServiceInput, ServiceOutput, Language } from "./types"
...
+    
+    if (cargo) {
+      let tarBuffer = new Tar();
+
+      let files = input.files;
+      Object.entries(files).forEach(
+        ([name, file]) => {
+          tarBuffer.append(name, file.content, {})
+        }
+      );
```
tschneidereit: This could just use a `for .. of` loop:
```js
for (const [name, file] of Object.entries(files)) {
  tarBuffer.append(name, file.content, {});
}
```

src/@types/tar-js/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+ *~ For example, if you were writing a file for "super-greeter", this
+ *~ file should be 'super-greeter/index.d.ts'
+ */
+
+/*~ Note that ES6 modules cannot directly export class objects.
+ *~ This file should be imported using the CommonJS-style:
+ *~   import x = require('someLibrary');
+ *~
+ *~ Refer to the documentation to understand common
+ *~ workarounds for this limitation of ES6 modules.
```
tschneidereit: I assume you just copied this over from tar-js, so this is more a "huh" comment than anything else, but: ES6 modules can export class objects perfectly well 🤷‍♀️

tschneidereit:
Oh, and the travis build should be passing, of course :)

src/@types/tar-js/index.d.ts
skade: Yeah, I forgot cleanup there. This is from the TypeScript Docs: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html

src/service.ts
skade: Thanks, change it in both places.

src/service.ts
```diff
@@ -240,6 +240,76 @@ export class Service {
...
+    for (const { name, item } of resultItemsIterable) {
+      const { content } = item;
+      if (content) {
+        outputFiles[name] = content;
+      }
+    }
+    return outputFiles;
+  }
+
+  static async compileFilesRecursive(files: File[], from: Language, to: Language, options = ""): Promise<{ [name: string]: (string|ArrayBuffer); }> {
```
yurydelendik: Let's remove this method -- user will specify all files in the array. Create an issue to create "find" the files by pattern (that will be feed into the compileFiles)

yurydelendik:
I don't see a way to create "Cargo.toml" -- we need to add this functionality to test this PR.

yurydelendik:
> I don't see a way to create "Cargo.toml"

Added in 7f0b0270c1e2f70480f4a18cb6e46a5ae2df0c17

skade:
@tschneidereit I made the build passing and refactored it as @yurydelendik asked for.

src/service.ts
tschneidereit: Can we do this later? We really need to get this landed ASAP.

skade: compileFiles currently flattens all files. (`Cargo.toml`, `src/lib.rs` will become `Cargo.toml`, `lib.rs`).

yurydelendik:
Thank you for the patch

PR #255, opened 2018-05-31
Revert Arc template row/col switch that will be obviated by a fix in the Arch website

tschneidereit:


tschneidereit:
Only affects the Arc embedding, merging without review.

PR #271, opened 2018-06-02
fix(arc): improve rust template

ashleygwilliams:


tschneidereit:
This only affects Arc templates, merging without detailed review.

PR #281, opened 2018-06-04
Remove forceUpdate from src/components/UploadFileDialog.tsx

ollelauribostrom:
Associated Issue: #91 

### Summary of Changes
* Removed forceUpdate from UploadFileDialog.tsx + did some general refactoring
* Added the convenience method Directory.hasChildren + tests

### DIFF
--- a/tests/components/UploadFileDialog/UploadFileDialog.spec.tsx
+++ b/tests/components/UploadFileDialog/UploadFileDialog.spec.tsx
@@ -40,20 +40,16 @@ describe("Tests for UploadFileDialog", () => {
       wrapper,
       addFiles() {
         const instance = wrapper.instance() as UploadFileDialog;
-        const root = instance.root.getModel();
-        const fileA = root.newFile("fileA", FileType.JavaScript);
-        const fileB = root.newFile("fileB", FileType.JavaScript);
-        instance.root = ModelRef.getRef(root);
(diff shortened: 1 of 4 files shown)


tests/components/UploadFileDialog/UploadFileDialog.spec.tsx
```diff
@@ -40,20 +40,16 @@ describe("Tests for UploadFileDialog", () => {
...
+        const fileA = instance.root.newFile("fileA", FileType.JavaScript);
+        const fileB = instance.root.newFile("fileB", FileType.JavaScript);
         wrapper.update();
         return [fileA, fileB];
       },
       addDirectory(name) {
         const instance = wrapper.instance() as UploadFileDialog;
-        const root = instance.root.getModel();
-        const directory = root.newDirectory(name);
+        const directory = instance.root.newDirectory(name);
```
yurydelendik: ModelRef.getRef and getModel were introduced to "track" direct mutations to the model. Looks like the `const directory = root.newDirectory(name);` code did just that. I recommend to extract this mutation from the component (and continue using model refs until we weed out all such instances)

ollelauribostrom:
@yurydelendik Not entirely sure if I understood your comment correctly but i've reverted back to using `root: ModelRef<Directory>;` in the component. Will this do the trick?

PR #282, opened 2018-06-04
Ignoring Typescript declaration files when collecting coverage

ollelauribostrom:
This removes the stack trace that gets printed when jest tries to collect coverage from src/@types/tar-js/index.d.ts

### Summary of Changes
* Ignoring TypeScript declaration files when collecting coverage


yurydelendik:
thanks~

PR #287, opened 2018-06-05
Fixing issue that makes it impossible to move file/folder

ollelauribostrom:
It seems like its currently not possible to move files/folders around in the directory tree. This is blocked somewhere around [here](https://github.com/wasdk/WebAssemblyStudio/blob/1c3104cf4c781a3480360f0e2239241a61b11bb5/src/components/DirectoryTree.tsx#L283). This PR adds a check to allow both drag and drop uploads and regular move actions.

### Summary of Changes
* Checking whether or not the drag data contains any elements (In other words - is it a file dragged from outside the browser or is it an element being moved?)

### Test Plan
- Create empty C project
- Verify that files in the directory tree can be dragged/dropped
- Verify that files can be uploaded via drag n drop
- Verify that this works in Firefox as well

### DIFF
--- a/src/components/DirectoryTree.tsx
+++ b/src/components/DirectoryTree.tsx
@@ -28,7 +28,7 @@ import { MonacoUtils } from "../monaco-utils";
 import { ViewTabs } from "./editor";
 import { ViewType } from "./editor/View";
 import { openFile, pushStatus, popStatus, logLn } from "../actions/AppActions";
-import { uploadFilesToDirectory } from "../util";
+import { uploadFilesToDirectory, isUploadAllowedForMimeType } from "../util";
 
 export interface DirectoryTreeProps {
   directory: ModelRef<Directory>;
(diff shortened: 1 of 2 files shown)


src/components/DirectoryTree.tsx
```diff
@@ -281,24 +281,27 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
...
       onDragOver(tree: ITree, data: IDragAndDropData, targetElement: File, originalEvent: DragMouseEvent): IDragOverReaction {
-        const items = Array.from(originalEvent.browserEvent.dataTransfer.items);
-        function mimeTypeUploadIsAllowed(type: string) {
-          console.log(type);
-          if (type === "") { // Firefox doesn't show the "application/wasm" mime type.
-            return true;
+        // File being dragged into the browser
+        if (!(data as any).elements) {
+          const items = Array.from(originalEvent.browserEvent.dataTransfer.items);
+          function mimeTypeUploadIsAllowed(type: string) {
```
yurydelendik: Normally (utility) function definition (without closure) needs to be done outside other function/class definition. Can you move it outside class method. Can we also fix bad name to be in form of verb-subject, e.g. `isUploadAllowedForMimeType(type: string)`?

src/components/DirectoryTree.tsx
```diff
@@ -281,24 +281,27 @@ export class DirectoryTree extends React.Component<DirectoryTreeProps, {
...
+            if (type === "") { // Firefox doesn't show the "application/wasm" mime type.
+              return true;
+            }
+            return fileTypeForMimeType(type) !== FileType.Unknown;
+          }
+          // In Firefox, tree elements get data transfer items with the "text/uri-list" type. This is a
+          // workaround to ignore that behavior.
+          const firefoxWorkaround = !items.find(item => item.type === "text/uri-list");
```
yurydelendik: nit: `firefoxWorkaround` a bad name -- I suggest `hasItemsUriListType`. Also usage of `items.some` or `items.every` instead of `items.find` will make code more readable.

yurydelendik:
Thanks!

PR #293, opened 2018-06-11
Adding tests for EventDispatcher class

ollelauribostrom:
### Summary of Changes
* Added tests for the EventDispatcher class (in model.ts)


yurydelendik:
Thank you for the patch.

PR #294, opened 2018-06-11
Adding tests for Problem class

ollelauribostrom:
### Summary of Changes
* Added tests for the Problem class (in model.ts)


yurydelendik:
Thank you for the patch

PR #296, opened 2018-06-12
Removing nested classes and adding tests (Controller, DnD, Template)

ollelauribostrom:
This makes it easier to test both the concerned components and the classes that previously were nested inside of them.

### Summary of Changes
* Broken out nested Controller and DnD classes from DirectoryTree.tsx & Problems.tsx to make them easier to test
* Moved FileTemplate and ProblemTemplate classes to seperate file (now inherits from shared Template class)
* Added tests for Controller, DnD and Template classes


yurydelendik:
Thank you for the patch.

PR #297, opened 2018-06-12
Listen to onDidChangeProblems in ControlCenter.tsx

ollelauribostrom:
The Problems tab in the ControlCenter does not update on new problems (in other words, it gives no indication that there is a problem). This PR fixes this by registering a listener for onDidChangeProblems events when the ControlCenter component mounts.

### Summary of Changes
* Listen to onDidChangeProblems in ControlCenter.tsx

### Test Plan
- Create empty C project
- Open main.c
- Change 42 to a string
- Save 
- Build
- The Problems tab should now indicate that there is one problem


yurydelendik:
Thank you.

Issue #288, opened 2018-06-05
Don't expose all actions in upload file/dir dialog


yurydelendik:
Currently, the upload file and directory dialog uses standard content menu:

<img width="472" alt="screen shot 2018-06-05 at 10 02 00 am" src="https://user-images.githubusercontent.com/1523410/40984486-a367f6cc-68a7-11e8-9f33-c4e4a45825ad.png">

IMHO, it is not expected to have *all* actions in this content menu. The "rename", "delete" might be just enough for this functionality

PR #307, opened 2018-06-19
Limiting the amount of actions in upload file/dir dialog

ollelauribostrom:
Associated Issue: #288 

### Summary of Changes
* Added a new prop to DirectoryTree.tsx (makes it possible to only show upload related actions)
* Updated UploadFileDialog.tsx (passing the new prop and adding edit file functionality)
* Added two test cases

### Test Plan
- Create C "hello, world" project
- Open the upload dialog
- Upload some files
- Verify that you now only see the "edit" and "delete" actions when opening the context menu


yurydelendik:
Thank you for the patch.

Issue #283, opened 2018-06-05
Error: Should not be in a dragging state


brennie:
I dragged the pane between the console output and rendered output after I clicked "build and run" and triggered a fatal error:

```
Error: Should not be in a dragging state.
Stack trace:
C@https://webassembly.studio/dist/main.bundle.js:1:12968
setupSolver@https://webassembly.studio/dist/main.bundle.js:13:33037
componentWillReceiveProps@https://webassembly.studio/dist/main.bundle.js:13:32115
Mf@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:76:230
wh@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:100:228
kg@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:120:88
lg@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:120:384
fc@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:127:242
vb@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:126:226
ub@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:126:61
vd@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:124:438
ra@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:123:317
enqueueSetState@https://unpkg.com/react-dom@16/umd/react-dom.production.min.js:187:274
q.prototype.setState@https://unpkg.com/react@16/umd/react.production.min.js:18:428
bindAppStoreEvents/<@https://webassembly.studio/dist/main.bundle.js:13:81599
dispatch/<@https://webassembly.studio/dist/main.bundle.js:1:5012
dispatch@https://webassembly.studio/dist/main.bundle.js:1:4985
bindProject/<@https://webassembly.studio/dist/main.bundle.js:13:17504
dispatch/<@https://webassembly.studio/dist/main.bundle.js:1:5012
dispatch@https://webassembly.studio/dist/main.bundle.js:1:4985
popStatus@https://webassembly.studio/dist/main.bundle.js:1:10543
handleActions@https://webassembly.studio/dist/main.bundle.js:13:21769
@https://webassembly.studio/dist/main.bundle.js:13:21882
r</e.prototype._invokeCallback@https://webassembly.studio/dist/main.bundle.js:24:63864
[...]

PR #308, opened 2018-06-19
Fix bug in Split.tsx: Should not be in a dragging state

ollelauribostrom:
Associated Issue: #283

### Summary of Changes
* Aborting any ongoing drag when the Split component receives new props
* Adding a test case

### Test Plan
- Create C "hello, world" project
- Drag a split
- Build the project (cmd + b) while still dragging the split
- It should no longer throw an error as described in #283


yurydelendik:
Thanks

PR #310, opened 2018-06-20
Extract models from model.ts

ollelauribostrom:
### Summary of Changes
* Extracting all models from model.ts to own files in /models
* Updating all imports


yurydelendik:
Looks good. Thank you for the patch.

Issue #131, opened 2018-02-23
Upload directories


yurydelendik:
See https://github.com/wasdk/WebAssemblyStudio/pull/36#issuecomment-363899987

We need to investigate if it's possible to upload a directory. It will provide alternative to have templates locally and upload them into WebAssembly Studio.

yurydelendik:
See https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory

haxxorsid:
It will be helpful if someone first works on #39 and also adds drag, drop functionality.
Then this can work out a bit easier.

mbebenita:
Partially fixed by #154, there's sill some follow up work to be done here, like error checking and dragging and dropping.

PR #311, opened 2018-06-20
Implementing drag and drop upload of directories

ollelauribostrom:
Associated Issue: #131

### Summary of Changes
* Using DataTransferItemList (dataTransfer.items) instead of FileList (dataTransfer.files)
* Adding a function to recursively read an uploaded directory

### Test Plan
- Upload a folder via drag and drop
- Verify that this works in both Firefox and Chrome
- It should now also be possible to move files around in the upload file/dir dialog


yurydelendik:
Thank you for the patch

PR #329, opened 2018-07-04
Fix sourcemapping in coverage reports

ollelauribostrom:
Jest does not correctly map lines in coverage reports back to the original Typescript files. This PR fixes this by generating inline source maps during test-runs which is picked up by Jest.

### Summary of Changes
* Adding the inlineSourceMap option to tsconfig.test.json
* Adding the --no-cache flag when collecting coverage with Jest


yurydelendik:
Thanks

Issue #301, opened 2018-06-13
Empty context menu in the Problems view


ollelauribostrom:
I am getting an empty context menu in the Problems view. It looks like this:

<img width="636" alt="empty-context" src="https://user-images.githubusercontent.com/16004130/41373904-efa4913e-6f51-11e8-99b0-33f44d761819.png">

Steps to reproduce:
- Create empty C project
- Change the return in main.c from 42 to a string
- Save & Run build
- Open the Problems view
- Right click to open the context menu

What is the expected behavior here? Should there be any actions or can we disable opening the context menu if there aren't any associated actions?

ollelauribostrom:
This also happens in the upload file/dir dialog

PR #331, opened 2018-07-05
Hiding the context menu if no actions are provided

ollelauribostrom:
Associated Issue: #301

### Summary of Changes
* Hiding the context menu if no actions are provided (in src/monaco-controller.ts)
* Updating/adding tests

### Test Plan
- Open the Problems view
- Right click to open the context menu
- You should no longer get an empty context menu (as shown in #301)


yurydelendik:
Thank you for the patch

Issue #286, opened 2018-06-05
Handle file collisions on upload


ollelauribostrom:
When uploading a file/directory with an already existing name, an error is currently thrown (at Project.addFile). This collision needs to be handled (starting point in the [App Component](https://github.com/wasdk/WebAssemblyStudio/blob/1c3104cf4c781a3480360f0e2239241a61b11bb5/src/components/App.tsx#L684)). 

Somewhat related to #172 where its mentioned that it could be a good idea to rename dragged/uploaded files to main.c.2 etc when a collision occurs (instead of preventing the drag/throwing this error).

yurydelendik:
Re-opening.. looks like we regressed some functionality at #335

PR #332, opened 2018-07-05
Improving the handling of name collisions

ollelauribostrom:
Associated Issue: #286 

### Summary of Changes
This improves the handling of name collisions on file/directory move and upload. Instead of preventing the drop, files/directories are renamed to `main.c.2`, `src.2` etc. as proposed in #172

* Name collisions are now handled in Directory.ts when adding a new file
* The DragAndDrop class now accepts dropping files even if there are a name collision
* Tests & upload utilities have been updated

#### Drag
- Create empty C project
- Create a new file called main.c
- Drag the new main.c into the src directory
- The file should be renamed to main.c.2

#### Upload
- Upload a file/directory
- Upload the same file/directory again
- The second upload should be renamed

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -681,7 +681,6 @@ export class App extends React.Component<AppProps, AppState> {
             this.setState({ uploadFileDialogDirectory: null });
           }}
           onUpload={(files: File[]) => {
-            // TODO: Handle file collisions.
             files.map((file: File) => {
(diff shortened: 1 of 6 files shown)


yurydelendik:
We are dealing too much with extensions to determine e.g. file type. It will be better to add suffix before the known(?) extension, e.g. 'file.js' -> 'file.2.js' (or 'file (2).js')

yurydelendik:
Thank you

Issue #335, opened 2018-07-09
Each build creates a new WASM file


d07RiV:
I think this only started happening recently - every time the code is built, a new file (named `main.3.wasm`, `main.4.wasm`, etc) is created. This means we have to update `.js` every time to use the correct file, which gets pretty annoying.

Issue #333, opened 2018-07-05
File/directory upload via regular input not working


ollelauribostrom:
Regular file/directory upload has stopped working since #311.

### Possible fix
The onChange handler [here](https://github.com/wasdk/WebAssemblyStudio/blob/ef9c8ed421e82dbd765473a8d3aebd46d68dc1a1/src/components/UploadFileDialog.tsx#L92) needs to be changed back to `this.handleUpload(e.target.files);` and the `uploadFilesToDirectory` utility function needs to be updated to handle both DataTransferItemLists and FileLists

### Steps to reproduce:
- Create empty C project
- Open the upload file/dir dialog
- Try to upload a file/directory by clicking on the Files / Directory buttons

PR #342, opened 2018-07-19
Fix upload & improve the handling of name collisions

ollelauribostrom:
Associated Issue: #333, #286 

This is a slight rework of PR #334 and #332 combined that addresses the issue described in #335 (Each build creates a new WASM file).

### Summary of Changes
* Fix file/directory upload via regular input
* Handle name collisions when moving/uploading files (but not when building/assembling/disassembling etc)

#### Name collision (Drag)
- Create empty C project
- Create a new file called main.c
- Drag the new main.c into the src directory
- The file should be renamed

#### Name collision (Upload)
- Upload a file/directory
- Upload the same file/directory again
- The second upload should be renamed

#### Name collision (Build)
- Create empty C project
- Build it
- Build it again
- It should not generate more than one wasm file

#### Upload
- Create empty C project
- Upload a file/directory by clicking on the Files / Directory buttons
- Upload a file/directory via drag and drop
- Both regular upload and dnd should be working in FF and Chrome

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -681,7 +681,6 @@ export class App extends React.Component<AppProps, AppState> {
             this.setState({ uploadFileDialogDirectory: null });
           }}
           onUpload={(files: File[]) => {
-            // TODO: Handle file collisions.
             files.map((file: File) => {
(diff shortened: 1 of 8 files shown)


yurydelendik:
Thank you for the patch.

Issue #351, opened 2018-07-24
TypeError: Failed to fetch


yurydelendik:
https://sentry.io/webassembly-studio/webassembly-studio/issues/586192533/

```
TypeError: Failed to fetch
  at _promiseRejectionHandler (/)
```

yurydelendik:
https://sentry.io/share/issue/55f18d8ebfa44c89bd8377320f336982/

ollelauribostrom:
This happens when the user tries to run a project before building it. Trying to fetch the out/main.wasm file to run returns a 404 which then leads to 
- [AsyncCompile: Wasm decoding failed: expected magic word 00 61 73 6d, found 3c 21 44 4f @+0](https://sentry.io/share/issue/90944166881a4401b8ab48201d56d36e/) 
or 
- [Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.](https://sentry.io/share/issue/55f18d8ebfa44c89bd8377320f336982/) as linked above.

Steps to reproduce (fiddles):
- Visit https://webassembly.studio/?f=gvuw4enb3qk & Run it without building
- Visit https://webassembly.studio/?f=ivzzdwn7fcn & Run it without building
- The errors are uncaught (and logged to sentry)

Steps to reproduce (normal):
- Open https://webassembly.studio/
- Create empty C project & Run it without building
- The error is caught and logged to output console in the UI

Im not sure why these errors are not caught when loading the project from a fiddle. Any ideas? One possible overall solution/improvement could be to simply disable the "Run" button before a wasm file has been built?

iFlameing:
Hi,I worked on disabling the Run button before a wasm file .I disabled it but when I run the test I encountered with error 3 snapshots test failed(should render correctly (embedded)).What should I do know.

ollelauribostrom:
@iFlameing Seems likely that your changes might break a few tests. Open a pull request and I'll have a look!

ollelauribostrom:
The reason why this happens for the fiddles linked above is because they don't catch/handle errors thrown when fetching the wasm file from main.html/main.js. These Sentry issues can therefore be ignored/resolved since nothing is wrong with WebAssembly Studio. 

However, I still propose we improve the error messages when this happens. In my opinion it would be much more clear what has gone wrong if we logged something like:
`[error]: Failed to fetch: https://localhost:28443/out/main.wasm`

mollthecoder:
[...]

PR #357, opened 2018-08-03
Improve error message during sandbox run

iFlameing:
Associated Issue: #351 Failed to fetch

Disabling the Run button and when the project is build the Run button is enable .

I tested by Running locally "hello world"  project  in C.

Example test plan:

- [x] Click Build button 

### DIFF
--- a/src/components/Sandbox.tsx
+++ b/src/components/Sandbox.tsx
@@ -85,7 +85,7 @@ export class Sandbox extends React.Component<{}, {}>  {
       error.apply(contentWindow.console, arguments);
     })(contentWindow.console.error);
     // Hijack fetch
-    contentWindow.fetch = (input: string, init?: RequestInit) => {
+    contentWindow.fetch =  async (input: string, init?: RequestInit) => {
       const url = new URL(input, "http://example.org/src/main.html");
       const file = project.getFile(url.pathname.substr(1));
       if (file) {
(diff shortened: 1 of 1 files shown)


ollelauribostrom:
Been thinking some more about this. If we disable the Run button before a build has been performed, users can no longer upload and run projects that has already been built (Project already containing a .wasm file).

Maybe it would be better to have a look in [Sandbox.run](https://github.com/wasdk/WebAssemblyStudio/blob/f13b623849f6838332e7e1da96840bdac4636f69/src/components/Sandbox.tsx#L88) to fix error messages when fetch fails. In my opinion it would be much more clear what has gone wrong if we logged something like:

`[error]: Failed to fetch: https://localhost:28443/out/main.wasm`

src/components/Sandbox.tsx
```diff
@@ -82,6 +82,7 @@ export class Sandbox extends React.Component<{}, {}>  {
     })(contentWindow.console.warn);
     contentWindow.console.error = (error => function() {
       logger.logLn(Array.prototype.join.call(arguments), "error");
+      logger.logLn(Array.prototype.join.call(["Failed to fetch:https://localhost:28843/out/main.wsm"]),"error");
```
ollelauribostrom: This will make all errors say "Failed to fetch:https://localhost:28843/out/main.wsm" no matter what has gone wrong, which is not what we want.

iFlameing: ok I am working on it as you suggested.

iFlameing:
@ollelauribostrom   I made all the changes as you said.Is it ok or I have to modify something .

yurydelendik:
Thank you for the patch

Issue #298, opened 2018-06-13
support for C headers


kattkieru:
Hi there,

Perhaps I've missed it, but it seems as if there's no way to add a header (.h, .hpp) file to the project.  It would be great to have that functionality added.

Thanks!

yurydelendik:
Currently the service allows to do that, see https://github.com/yurydelendik/clang-heroku-slug/blob/master/app/src/build.js#L11. We need to implement [ClangService](https://github.com/wasdk/WebAssemblyStudio/blob/2101a6da88e5297275b2cf7d8180690a0f6b7bef/src/compilerServices/clangService.ts) to send all files there.

iFlameing:
@yurydelendik  How I can fetch the header file with the extension .h and .hpp and send to compilerservice because right now we can only fetch .c and .cpp and the other extension not .h or .hpp extension and how I can attach the main.c and the header file , can you elaborate please!.

yurydelendik:
@iFlameing the .h/.hpp shall be sent to the service in the same way as .c/.cpp. If wasm-studio does not support .h extension -- we shall add it. We may want to mark these with different type, rather than  C or CPP, so they will not be included in clang arguments.

iFlameing:
if we send the .h/.hpp file same way then we have to define a new clang service because createCompilerService does not support the h/hpp file . I try your approach of sending the header.h file as same as c/cpp but it gives me error "[error]: createCompilerService: not supported h->wasm".
How I can remove this error.

PR #365, opened 2018-09-04
Now We can Use Header File .

iFlameing:
Associated Issue: #298 

### Summary of Changes

* change 1
I changes the service.ts, compilerservice.ts and clangservice.ts according to need.
 * change 2
I made some modification in build.ts and in the template also for simple implementation of the header file.
### Test Plan
Due to some  change in the service.ts and specially in clangservice.ts some testcase fail.
But I check the implementation of header file by manually checking .

Example test plan:

- [x] Click Build button 

screenshot from 2018-09-04 21-27-42

screenshot from 2018-09-04 21-27-33


AnshulMalik:
This looks really good, I'll review in a moment.

yw662:
So can we also compile multiple c/cpp files if this is merged ?

AnshulMalik:
@iFlameing after looking into it, it seems like we are just handling a specific case of a single header file. Its a great start, I was hoping if we can come up with some more generic solution for this, so that we can compile as many header files as we want.

iFlameing:
@AnshulMalik  The main problem regarding the multiple file is that, we have to mention the file name into the build.ts  file manually.

AnshulMalik:
Can we extract all the includes somehow from the one source file we are compiling. 
And the automatically add those to the list of files to be compiled?

iFlameing:
@AnshulMalik This might be possible but for compiling we have to send all the file to server and if we include single file in main.c and import other file into header.h then we also have to send all the file which we are importing into the header.h .

AnshulMalik:
So it's recursive then, don't you love recursion? :P

iFlameing:
@AnshulMalik  Now it is supporting for multiple header file .

templates/hello_world_c/build.ts
```diff
@@ -2,7 +2,7 @@ import * as gulp from "gulp";
 import { Service, project } from "@wasm/studio-utils";
 
 gulp.task("build", async () => {
-  const data = await Service.compileFile(project.getFile("src/main.c"), "c", "wasm", "-g -O3");
+  const data = await Service.compileFile([project.getFile("src/main.c"),project.getFile("src/header.h")], "c", "wasm", "-g -O3");
```
yurydelendik: use Service.compileFiles instead

src/service.ts
```diff
@@ -197,18 +198,17 @@ export class Service {
     return annotations;
   }
 
-  static async compileFiles(files: File[], from: Language, to: Language, options = ""): Promise<{ [name: string]: (string|ArrayBuffer); }> {
+  static async compileFiles(files: Array<any>, from: Language, to: Language, options = ""): Promise<{ [name: string]: (string|ArrayBuffer); }> {
```
yurydelendik: Not sure why we are changing File[] to File[][] ?

src/service.ts
```diff
@@ -61,7 +61,8 @@ export { Language } from "./compilerServices";
 
 function getProjectFilePath(file: File): string {
   const project = file.getProject();
-  return file.getPath(project);
+  const result = file.getPath(project);
```
yurydelendik: Unneeded change

src/compilerServices/types.ts
```diff
@@ -30,7 +30,9 @@ export enum Language {
   JavaScript = "javascript",
   TypeScript = "typescript",
   Toml = "toml",
-  Text = "text"
+  Text = "text",
+  h = "h",
```
yurydelendik: Capitilize H and Hpp

src/compilerServices/clangService.ts
```diff
@@ -31,23 +31,26 @@ export class ClangService implements CompilerService {
...
-    }
     const [ fileRef ] = Object.keys(input.files);
-    const src = files[0].content;
     const from = this.lang;
+    const inputFile=Object.keys(input.files);
+    const File = function(inputFile:any,file:any){
+      const compileFile =[];
+      for(let i=0;i<inputFile.length;i++){
+        const f={
+          type:inputFile[i].split("/")[1].split(".")[1],
```
yurydelendik: type can defined on the InputFile

src/compilerServices/clangService.ts
```diff
@@ -31,23 +31,26 @@ export class ClangService implements CompilerService {
 
   async compile(input: ServiceInput): Promise<ServiceOutput> {
     const files = Object.values(input.files);
-    if (files.length !== 1) {
-      throw new Error(`Supporting compilation of a single file, but ${files.length} file(s) found`);
-    }
     const [ fileRef ] = Object.keys(input.files);
-    const src = files[0].content;
     const from = this.lang;
+    const inputFile=Object.keys(input.files);
```
yurydelendik: very poor code formatting and identifiers names -- hard to read the change

src/service.ts
iFlameing: This is all happen just because of  I am not using the service.compileFiles  I made the changes based on the service.compileFile.

src/service.ts
iFlameing: I revert the change.

src/compilerServices/types.ts
iFlameing: I change it.

src/compilerServices/clangService.ts
iFlameing: I should be careful to identifiers names and formatting the codes.

src/compilerServices/clangService.ts
iFlameing: ok

PR #366, opened 2018-09-19
Test improvements

ollelauribostrom:
General improvements of the test suite. 

### Summary of Changes
* Removed all hardcoded `await wait(10)` from the test suite (using the wait-until-promise package instead).
* Making sure any console.error calls from React results in failing tests by spying on console.error in `test-setup.js` and throwing any logged errors. After doing this, a few tests were failing due to missing key properties in `App.tsx` but this is now fixed.
* A few tests in `Markdown.spec.tsx` were previously failing when running the test suite using Node 10. This is now fixed by using the wait-until-promise package.
* Updating Enzyme and enzyme-adapter-react-16 and removing the use of `.parent()` in `UploadFileDialog.spec.tsx` and `Workspace.spec.tsx` due to airbnb/enzyme#1780
* Updating Stryker and `stryker.conf.js`.
* Updating TypeScript and doing some updates to `util.ts` and `taskRunner.ts` to make the compiler happy again.
* Updating TSLint.
* Added the [allowSyntheticDefaultImports](https://www.typescriptlang.org/docs/handbook/compiler-options.html) in tsconfig.test.json to allow default import from wait-until-promise

### Example of using wait-until-promise in a test
Instead of
```js
await wait(50) // This always waits for 50ms
const newFile = root.getFile("file.js");
```
We can now do
```js
// Wait until the passed function returns a truthy value (i.e. the file has been created)
const newFile = await waitUntil(() => root.getFile("file.js")); 
```
We no longer have to wait for the full 50ms if the file is created quicker than that. In my opinion this makes it more clear what we are waiting for. If the function passed never returns a truthy value, or it times out, the promise is rejected which makes the test fail.


yurydelendik:
Thank you for the patch

Issue #348, opened 2018-07-24
[Discussion] E2E tests


ollelauribostrom:
As a result of my GSoC project this summer, test coverage is about to reach acceptable levels. To further improve the testing of WebAssembly Studio I would like to bring up the potential need of some E2E tests for some key functionality (like building/running etc). 

This would help verify that key functionality is working as expected, which individual unit tests or component integration tests might still not catch. Issue #335 is an example of something that would have be picked up by an E2E test. 

Since we are using Jest as our test runner, using [puppetter](https://github.com/GoogleChrome/puppeteer) / [jest-puppetter](https://github.com/smooth-code/jest-puppeteer) could be an option for this as described in the [Jest documentation](https://jestjs.io/docs/en/puppeteer). What are your thoughts? Would love some input/suggestions before putting together a proof of concept.

PR #367, opened 2018-09-20
Run e2e tests in Travis build

ollelauribostrom:
Associated Issue: #348 

All dependencies that are needed to run the e2e tests are already installed when running `npm install` in the Travis build. Why not also run the tests? (takes about 30s)

### Summary of Changes
* Added `npm run e2e` to the scripts in `.travis.yml`
* Fix paths in the `templates` script in `package.json`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ install:
 - npm install
 script:
 - npm test
+- npm run e2e
 - npm run coveralls
 - npm run templates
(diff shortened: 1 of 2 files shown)


ollelauribostrom:
Not sure why the build fails, need to investigate this further. Consider this PR a WIP!

yurydelendik:
Thank you for the patch

Issue #262, opened 2018-06-01
Embed: Successful build outputs nothing


skade:
Just a nit, but a successful build outputs nothing in the console, which confused people.

Issue #156, opened 2018-03-03
Empty (sic) C project Build gives only one empty line in Output


chrisjj:
And no indication that the build took place.

Perhaps this line was supposed to be a completion message?

![image](https://user-images.githubusercontent.com/6157073/36938826-54e5d1a8-1f1f-11e8-92f5-59c4d2d4236a.png)

mbebenita:
The only indication at the moment is that a `main.wasm` file is created. Would be great to figure out a better way to give feedback.

haxxorsid:
@mbebenita Reminds me of: [this](https://github.com/wasdk/WebAssemblyStudio/issues/22#issuecomment-366009626)

chrisjj:
Call me old-fashioned :) but how about 

    [info]: Build complete

mbebenita:
That's a good suggestion. However, there are several other actions that users can take, through context menus and such which would need some kind of feedback too. We could write all these to the output window, but we haven't though through all the implications just yet.

PR #370, opened 2018-10-02
Giving Information to user When Build is finished

iFlameing:
Associated Issue: #262 and #156

### Summary of Changes

* Just adding a line in taskRunner.ts for showing the info .

### Test Plan

By manually testing on webAssembly editor.
Example test plan:

- [x] Click Build button 

### Screenshots/Videos (OPTIONAL
screenshot from 2018-10-02 15-00-49

### DIFF
--- a/src/utils/taskRunner.ts
+++ b/src/utils/taskRunner.ts
@@ -183,6 +183,7 @@ export async function runTask(
   if (gulp.hasTask(name)) {
     try {
       await gulp.run(name);
+      logLn("Build is Finished", "info");
     } catch (e) {
       logLn(e.message, "error");
     }
--- a/src/utils/taskRunner.ts
+++ b/src/utils/taskRunner.ts
@@ -183,7 +183,7 @@ export async function runTask(
   if (gulp.hasTask(name)) {
     try {
       await gulp.run(name);
-      logLn("Build is Finished", "info");
+      logLn(`Task ${name} is completed `, "info");
     } catch (e) {
       logLn(e.message, "error");
     }
--- a/src/actions/AppActions.ts
+++ b/src/actions/AppActions.ts
@@ -329,7 +329,7 @@ export async function run() {
 
 export async function build() {
   pushStatus("Building Project");
-  await runTask("default");
+  await runTask("build");
   popStatus();
 }
 


src/utils/taskRunner.ts
```diff
@@ -183,6 +183,7 @@ export async function runTask(
   if (gulp.hasTask(name)) {
     try {
       await gulp.run(name);
+      logLn("Build is Finished", "info");
```
yurydelendik: Let's change that to informational message that it will contains the task name, e.g. `Task ${name} is completed`. It will be probably a good idea to add `Task ${name} is running`.

iFlameing: ${name} is default Show the output is become is   [info]: Task default is completed  If i change or stick with previously because  [info]: Build is completed is  a good information for the user because default task is doing the build process.

yurydelendik: problem then: it will show on any task as build is completed, which is misleading more. Based on build.ts files we have "build" tasks.

iFlameing: No,It only show  when "build"  is finished .  **"default "  is not any task it is used as representation** whenever the default task run all the mention task will run but in **build.ts file there is only one mention task in default i.e "build"** whenever the " default " task is run it automatically run the "build"  task.And output the out/main.wasm file.

iFlameing: We can also do like in  place of default task we should pass  build task because both are the same. what you say.

iFlameing: > No,It only show when "build" is finished . **"default " is not any task it is used as representation** whenever the default task run all the mention task will run but in **build.ts file there is only one mention task in default i.e "build"** whenever the " default " task is run it automatically run the "build" task.And output the out/main.wasm file. There is not any misleading .We just run the build task as different manner.In place of default I also mention the "build " task and it worked.

iFlameing:
I just made another commit regarding the issue and now the problem is solved.

yurydelendik:
> I will also change the test file accordingly.

Yes, test(s) needs to be adjusted accordingly. 

> Coverage decreased (-0.02%) 

Also, it expected to add tests that check if output panel is indeed populated with message.

What if task/build failed?

yurydelendik:
Fixed by #382

iFlameing:
> Fixed by #382

No It is not fix by #382 that pr is for only for changing the test.
This pr is for outputting the build is finished.

iFlameing:
Sorry this pr is open just because of my confusion please close it.

PR #374, opened 2018-10-08
Improving build scripts

ollelauribostrom:
Some minor improvements to the CI build process. Instead of first running all of the tests, and then running them again to collect coverage we can just run them once. Removing the `--no-cache` flag should also speed things up a bit.

### Summary of Changes
* Updating the Travis config
* Adding `test:ci` script in package.json
* Removing `--no-cache` flag from coverage script in package.json


yurydelendik:
Thank you for the patch

PR #378, opened 2018-10-12
Use `textContent` instead of `innerText`

Korilakkuma:
### Summary of Changes

- Use `textContent` instead of `innerText`
- Because as `innerText` is aware of CSS styling, it will trigger a reflow, whereas `textContent` will not.

### Test Plan

- [x] Click Build button 


yurydelendik:
Thank you for the patch.

PR #388, opened 2018-11-19
Update AssemblyScript compiler location

dcodeIO:
This PR updates the dist file locations for the AssemblyScript compiler to use the latest releases from GitHub again. Looks like this changed in #384 due to RawGit shutting down soon, so this switches to jsDelivr for this purpose. As before, essentially moves the responsibility for keeping these files in sync and up to date to us on the AssemblyScript side.


yurydelendik:
Thanks

Issue #400, opened 2019-01-24
Corrupted base64 encoding


jeff1010322:
The padding for base64 encoding is adding one too many equal signs when the byte remainder is 1.  

https://github.com/wasdk/WebAssemblyStudio/blob/fbf7d4ef5e09b2798caf5137081ff21bcd184aa1/src/util.ts#L135

This is based on a gist that correctly appends 2 equal signs:

https://gist.github.com/jonleighton/958841#file-base64arraybuffer-js-L53

yurydelendik:
Closed by #401

PR #401, opened 2019-01-25
Fixed extra = for base64 encoding

arpit73:
Associated Issue: #400 

### Summary of Changes

Changes made in src/utils.ts
* Changed the number of "=" signs used for base64 encoding  when byte remainder is 1. 
Changed from 3 to 2.
* Modified the test file to accommodate for the changes above.

### Test Plan
Rewrote test by modifying the expected results in tests/utils/utils.spects.

All tests passing.


yurydelendik:
Thank you for the patch!

Issue #50, opened 2018-02-09
Possible future exploration: running RLS in the browser


jonathandturner:
Something we might want to explore on the Rust side is to run the RLS in the browser.  This would let them get all the same IDE support they'd have locally in the browser experience (like autocompletes, live error reporting, etc).

yurydelendik:
Related #41 and #49

jonathandturner:
Good question.  Maybe @fitzgen might know a good person to do it.

fitzgen:
I think @alexcrichton would be able to help out more, or know who could. Unfortunately, my rustc hacking experience is quite limited.

alexcrichton:
AFAIK there hasn't been much effort to compiling rustc itself to run in a browser, but it should certainly be possible!

PR #447, opened 2019-12-19
rust-analyzer integration

JasperDeSutter:
Associated Issue: #50

### Summary of Changes

* Ported over the WebAssembly demo code of rust-analyzer (available here: https://rust-analyzer.github.io/)
* Extended rust IDE functionality with the help of rust-analyzer
   things like go-to definition, find usages, hover, completions, code lenses...

Note that this slows down the whole webpack build by quite a big margin. Especially fresh builds could take minutes more. Maybe another approach could be taken to include the final .wasm binary?

### Test Plan

These changes have been tried manually by just trying stuff out in the hello-world rust demo.
Some parts work a bit less well than on the original rust-analyzer WASM demo because the monaco editor seems to be outdated. This could be fixed in a follow-up PR, as it might need some bigger changes throughout the codebase.

example of find usages:
ra-lsp-wasm-studio


JasperDeSutter:
I tried including the wasm file in the latest commit instead of the rust source code. It is available here: https://github.com/rust-analyzer/rust-analyzer.github.io/tree/master (passed through wasm-opt to reduce size to 2.7MB instead of 5+)

